PR #47, opened 2018-09-21
[issue-46] Add Versioning, Fix Travis

### LINKED ISSUE #46: Add versioning
Since we are about to open up the repo and make the first release. We need to add versioning to the project.

adrianmo:
- Fix #46
- Create `Makefile` with options to build, create Docker image, run tests, and format code.
- Add version package under `pkg/version` with a `Version` variable that will be assigned during the build process via ldflags.
- Version number located in `VERSION` file.
- Create `Dockerfile`. Since we are now assigning the version during the build process, we need to create the Docker images ourselves to use binary instead of using on the `operator-sdk` tool to build the operator and produce the Docker images. 
- Create `scripts/check_license.sh` to check that all Go files include the license header.
- Create `scripts/check_format.sh` to check that all Go files comply with `gofmt` code style.
- Add `-version` flag to show version and quit
- Update `.travis.yaml` file.


adrianmo:
ping @maddisondavid @mcgG

cmd/pravega-operator/main.go
```diff
@@ -2,25 +2,43 @@ package main
...
 
 func main() {
-	printVersion()
+	if printVersion {
+		fmt.Println("pravega-operator Version:", version.Version)
+		fmt.Println("Go Version:", runtime.Version())
+		fmt.Printf("Go OS/Arch: %s/%s\n", runtime.GOOS, runtime.GOARCH)
+		fmt.Printf("operator-sdk Version: %v", sdkVersion.Version)
+		os.Exit(0)
+	}
```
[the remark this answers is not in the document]
adrianmo: @mcgG Yes, I agree with you that printing the version and exiting doesn't make sense in a k8 deployment. However, I think it can be useful for troubleshooting cases, e.g., when you don't know what's the version of a running operator process. Furthermore, this is just the first flag, but we can add more flags in the future to customize the behavior of the operator, e.g., watch all namespaces, metrics listen port, etc.

You can pass these options to the operator like this:

```
...
spec:
      containers:
        - name: pravega-operator
          image: maddisondavid/pravega-operator:latest
          ports:
          - containerPort: 60000
            name: metrics
          command:
          - pravega-operator
          - -flag1
          - -flag2
          imagePullPolicy: Always
...
```

or if you run the operator locally:

```
$ operator-sdk up local --operator-flags "-version"
```

pkg/version/version.go
```diff
@@ -0,0 +1,15 @@
...
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package version
+
+var (
+	Version = "0.0.0-master"
```
fpj: I don't understand the version format here. Why `master` and why `0.0.0`? Should it be `0.1.0 `?

cmd/pravega-operator/main.go
```diff
@@ -2,25 +2,43 @@ package main
 
 import (
 	"context"
+	"flag"
```
fpj: What are these imports?

fpj:
@adrianmo note that the PR build has failed.

cmd/pravega-operator/main.go
maddisondavid: Just standard GO imports https://golang.org/pkg/flag/

pkg/version/version.go
adrianmo: @fpj this is just a placeholder for the version. The initial `0.0.0-master` just indicates that we have not released any version yet and we are based on master for the moment. For the first release, this field will be updated to `0.0.1`.

adrianmo: @maddisondavid thanks for the feedback. I'll explore that alternative.

adrianmo:
@maddisondavid @fpj @mcgG can you please review the PR again? I've made important changes to it. Check the edited description for details.

VERSION
```diff
@@ -0,0 +1 @@
+0.0.0-master
```
fpj: It sounds good to move the version here, but I'm still confused about how we are going format the release version. We need to agree on it and I'd rather have this PR reflecting such a format.

adrianmo: So the way other operators and the Operator SDK are versioning their projects is the following. The first release will be `0.0.1` because we still consider the project in an alpha status. Once released, the version is changed to `0.0.1+git` to indicate that there are changes on top of `0.0.1`. Then, `0.0.2` is released and the version immediately changed to `0.0.2+git`. This is the versioning strategy that similar projects are using and we plan to use.

fpj: I don't understand what the digits mean if you do. Bumping the last version is supposed to indicate that it is a bug-fix release, but what you are saying does not seem to indicate that.

https://semver.org

adrianmo: Makes sense... We'll then start with an initial version `0.1.0`. 

- `0`: indicates initial development. Anything may change at any time. The public API should not be considered stable.
- `1`: first set of features
- `0`: no bug fixes yet.

fpj: https://github.com/pravega/pravega-operator/issues/53

maddisondavid: This has been merged now, but this doesn't really solve the initial problem because although the version number is OUTSIDE the GO code, it's still part of the actual code in the repository.  A much better way that I see a lot of projects using now is to calculate the SemVer from GIT tags.

PR #55, opened 2018-09-28
Skip cleanup in Travis deploy stage

adrianmo:
Fix Travis deploy.

- Update Docker password
- run `make clean` as last build step.


.travis.yml
```diff
@@ -16,8 +16,10 @@ script:
 - make check
 - make test
 - make build
+- make clean
 
 deploy:
+  skip_cleanup: true
```
fpj: Why do you want to skip cleanup?

adrianmo: You are actually right. We don't need to skip clean since `make push` will automatically trigger all other make actions to build the app and image. I've removed that line.

PR #66, opened 2018-10-15
Issue 20: Added Helm Chart

### LINKED ISSUE #20: Add Helm Chart for installation
Develop a Helm chart for the operator.

maddisondavid:
Added Helm chart to deploy pravega-operator into Kubernetes cluster

Fix #20

Signed-off-by: David Maddison <maddisondavid@gmail.com>

### DIFF
--- a/charts/pravega-operator/Chart.yaml
+++ b/charts/pravega-operator/Chart.yaml
@@ -1,0 +1,14 @@
+name: pravega-operator
+version: 0.1.0
+appVersion: 2.6.0
+description: |
+  pravega operator deploys a custom resource for a pravega cluster, and a
+  pod to provision and scale pravega clusters.
+keywords:
+- pravega
+- storage
+home: https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator
+icon: https://avatars3.githubusercontent.com/u/25698199
+sources:
+- https://github.com/pravega/charts/blob/master/pravega-operator
+engine: gotpl
--- a/charts/pravega-operator/Chart.yaml
+++ b/charts/pravega-operator/Chart.yaml
@@ -1,6 +1,6 @@
 name: pravega-operator
 version: 0.1.0
-appVersion: 2.6.0
+appVersion: 0.1.0
 description: |
   pravega operator deploys a custom resource for a pravega cluster, and a
   pod to provision and scale pravega clusters.
--- a/charts/pravega-operator/templates/operator.yaml
+++ b/charts/pravega-operator/templates/operator.yaml
@@ -1,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "pravegaOp.fullname" . }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: {{ template "pravegaOp.fullname" . }}
+  template:
+    metadata:
+      labels:
+        name: {{ template "pravegaOp.fullname" . }}
+    spec:
+      containers:
+      - name: {{ template "pravegaOp.fullname" . }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        env:
+        - name: "WATCH_NAMESPACE"
+          value: "{{ .Values.watch.namespace }}"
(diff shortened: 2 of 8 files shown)


charts/pravega-operator/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+name: pravega-operator
+version: 0.0.1
```
adrianmo: The current pravega-operator version is `0.1.0`, not `0.0.1`.

maddisondavid:
Updated Chart Version.

Yes this is a copy of the one in `charts` repo and we should probably deprecate and remove the `charts` repo at some point as the charts should be with the components themselves (the charts repo was created a lot earlier when we were still figuring out our Helm strategy)

adrianmo:
@maddisondavid it looks like other projects are migrating their Charts to the official repository, which is kind of a catalog. Have we considered this option? 

https://github.com/helm/charts

charts/pravega-operator/templates/rbac.yaml
```diff
@@ -0,0 +1,62 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: flink-operator
```
EronWright: erroneous "flink-operator" here and in the role binding.

charts/pravega-operator/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+name: pravega-operator
+version: 0.1.0
+appVersion: 2.6.0
```
EronWright: The `appVersion` is supposed to indicate the version of the Pravega Operator that this chart deploys.

charts/pravega-operator/templates/operator.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pravega-operator
```
EronWright: I think we should parameterize the metadata as is typical of charts.   `{{ReleaseName}}`  type stuff.

charts/pravega-operator/values.yaml
```diff
@@ -0,0 +1,7 @@
+image:
+  repository: pravega/pravega-operator
+  tag: 0.1.0
+  pullPolicy: Always
+
+watch:
```
EronWright: How about we not expose this `WATCH_NAMESPACE` facility for now.  It would imply a different RBAC manifest (e.g. `Role` rather than `ClusterRole`).

charts/pravega-operator/Chart.yaml
adrianmo: I think `appVersion` indicates the version of Pravega, not the version of the Pravega Operator.

charts/pravega-operator/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: |
+  pravega operator deploys a custom resource for a pravega cluster, and a
+  pod to provision and scale pravega clusters.
+keywords:
+- pravega
+- storage
+home: https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator
+icon: https://avatars3.githubusercontent.com/u/25698199
```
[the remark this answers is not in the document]
adrianmo: It does work for me. Maybe you are behind the corp firewall?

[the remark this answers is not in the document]
adrianmo: However, the image doesn't look like an icon. But we can deal with it in a separate PR.

charts/pravega-operator/Chart.yaml
maddisondavid: `appVersion` can't specify what version of Pravega it is because that's something the user specifies when they provide the Pravega Image information as part of the `PravegaCluster`

charts/pravega-operator/values.yaml
maddisondavid: I've added multiple RBAC templates that trigger depending if the Watch namespace is set or not

charts/pravega-operator/Chart.yaml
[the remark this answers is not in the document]
maddisondavid: I can't find a Pravega icon.. how about this? http://blog.pravega.io/wp-content/uploads/2018/10/pr_logo3.png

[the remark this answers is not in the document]
adrianmo: Looks good to me.

adrianmo:
@EronWright can you give it a second look?

PR #67, opened 2018-10-16
Update VERSION to 0.1.0+git

adrianmo:


adrianmo:
@maddisondavid The `+git` suffix indicates that it's a _dirty_ version, i.e., there are some changes on top of the released version. I've seen that used in other operators, even in the Operator SDK project.

- https://github.com/operator-framework/operator-sdk/blob/master/version/version.go
- https://github.com/nats-io/nats-operator/blob/master/version/version.go
- https://github.com/deepinsight/mxnet-operator/blob/1.7/version/version.go
- https://github.com/coreos/etcd-operator/blob/master/version/version.go

However, I'm open to other approaches. What are you suggesting exactly?

maddisondavid:
The problem I guess is it's not a 'dirty' version if it's been committed.  From my understanding an dirty version is a GIT repo that has uncommitted modifications in it.

I guess this is a problem with using a version file embedded in the code itself.  Instead we could calculate the version during build time from GIT tags.  This has the advantage of being able to modify the product version without changing the code AND has the dual purpose of marking the source code with the version at the same time as setting that product version.

adrianmo:
Thanks for the insight @maddisondavid. After the discussion we had offline I'm now in favour of getting rid of the VERSION file and using Git information to version the operator. I'll submit a PR with the changes.

PR #70, opened 2018-10-17
Issue 53: Enhance project versioning

### LINKED ISSUE #53: Operator versioning strategy
The operator will use SemVer for versioning. The initial release will be `0.1.0`, with the following meaning:

- `0`: indicates initial development. Anything may change at any time. The public API should not be considered stable.
- `1`: first set of features.
- `0`: no bug fixes yet.

After each release, we will immediately append the `+git` suffix to the version number to indicate that there are changes on top of that version. E.g. after the first release, we will change the version to `0.1.0+git`. This is what the Operator SDK and other operators are also doing.

adrianmo:
- Base versioning on Git information instead of the external VERSION file.
- If the commits points to a tag, then only the tag is shown (e.g. `0.1.0`).
- Otherwise, it suffixes the tag name with the number of additional commits on top of the most recent tag (e.g. `0.1.0-4` if there are four commits on top of the tag `0.1.0`).
- If the working tree has local modification `-dirty` is appended to the version (e.g. `0.1.0-4-dirty`).
- `-version` shows also the Git commit SHA.
- Fix #53

Signed-off-by: Adrian Moreno <adrian@morenomartinez.com>


EronWright:
LGTM

PR #76, opened 2018-10-24
Issue 74: Handle Go errors correctly

finiks:
The Go errors in the `ReconcilePravegaCluster()` function were never handled correctly. That is, upon receiving an error by one of the called functions, the `err` variable would never be set.

This PR fixes this by explicitly setting the `err` variable every time a new call is made.

Signed-off-by: Olivier Bourgeois <olivier.bourgeois@emc.com>


finiks:
I've noticed while resolving this bug that only 1/3 of the "components" uses `logrus` to print errors. Should we use it everywhere? Nowhere? It seems a bit inconsistent at the moment.

adrianmo:
About the `logrus`usage... my opinion is that we should use everywhere.

PR #77, opened 2018-10-29
Issue 61: Support external connectivity

### LINKED ISSUE #61: Support external connectivity
# Overview
The Pravega clusters that are produced by the operator should support external connectivity (i.e. connectivity from outside the Kubernetes cluster).   The specific endpoints in question are the controller RPC/REST ports, and the segment store RPC port.

## Challenges
Pravega ingests data directly from client to a dynamic set of segment stores, unlike a conventional service that relies on a stable, load-balanced endpoint.   The client discovers the segment stores with the help of the controller, who's aware of active segment stores and their endpoint addresses.   Specific challenges include:
- advertising usable addresses to the client
- facilitating transport encryption (TLS) to the segment store (e.g. supporting hostname verification)
- optimizing internal connectivity vs external connectivity (e.g. avoiding an expensive route when possible)

### Vendor Specifics
- PKS: has option to use NSXT for Ingress.  Istio is apparently on the roadmap.
- GKE: see references at bottom

## Implementation
For conventional services, external connectivity is generally accomplished with an `Ingress` resource.   Ingress primarily supports HTTP(s) and it is unclear whether gRPC (which is HTTP/2-based) is supported ([ref](https://hackernoon.com/reviewing-grpc-on-kubernetes-8a705b928abd)).   

Ingress is probably not suitable for exposing the segment store.   For workloads that are similar to Pravega, e.g. Kafka, the typical solution is to use a `NodePort`.

Keep in mind that `Ingress` and services of type `LoadBalancer` may incur additional costs in cloud environments ([GCP pricing](https://cloud.google.com/compute/pricing#lb)).

### Multiple Advertised Addresses
Certain Pravega clients will be internal to the cluster, others external.  Imagine that the segment store advertised only an external address (backed by a `NodePort` or other type of service); would the performance of internal clients suffer due to a needlessly expensive route?   A mitigation would be to introduce support for numerous advertised advertised addresses ("internal"/"external").  Given a prioritized list, the client could strive to connect to the cheapest endpoint.  

This idea could extend to full-fledged multi-homing, where the segment store binds to a separate interface/port per endpoint, possibly with a separate SSL configuration per endpoint.

### NodePort Details
Be sure to set the `externalTrafficPolicy` field of the `Service` to `local`.  This will ensure that traffic entering a given VM will be routed to the segment store on that same VM.

One limitation of `NodePort` is that only a single segment store may be scheduled on a given cluster node.  If multiple were to be scheduled, some would fail with a port-unavailable error.  One way to avoid this is to use a [`DaemonSet`](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) to manage the segment store pods.

 ### References
- https://kubernetes.github.io/ingress-nginx/deploy/baremetal/
- https://hackernoon.com/reviewing-grpc-on-kubernetes-8a705b928abd
- https://github.com/kubernetes/kubernetes/issues/23291
- https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer
- [GKE Ingress](https://github.com/kubernetes/ingress-gce)
- [GKE Ingress HTTP/2 Support](https://github.com/kubernetes/ingress-gce/pull/146)

adrianmo:
## Change log description

This PR adds support for external connectivity (i.e. connectivity from outside the Kubernetes cluster). Below is a detailed description of the changes made and the reason why they were made.

#### Create a headless service for the Segmentstore set
As with the Bookies, Segmentstores are managed by a `StatefulSet`. Headless services provide DNS records for the Segmentstore set that can be used for intra-cluster communication, i.e., segmentstore pods can be access using the `<statefulset_name>-<ordinal>.<headless_service_name>` hostname (e.g. `pravega-segmentstore-1.pravega-segmentstore-headless` would resolve to the IP of second pod in the set).

#### Create an external service per Segmentstore pod
Pravega clients need to be able to directly reach every Segmentstore instance. We found multiple ways to achieve this:
- Using `NodePort` services. `NodePort` services allocate a free port (within the 30000-32767 range) in each Kubernetes node and forwards the traffic to the backend pod. However, this requires extra configuration on the external networking layer to allocate IP addresses, and allow and forward external traffic to the service. In addition, `NodePort` services are not yet supported on PKS w/ vSphere and NSX-T, which is our main platform.
- Using the `Ingress` resource. Ingress configures an HTTP server with routing rules to allow external communication with a service. However, most Ingress implementations do not support non-HTTP traffic, making it not an option for our HTTP2-gRPC traffic.
- Using `LoadBalancer` services. A `LoadBalancer` service configures an external load balancer (e.g. NSX-T in our platform) to allow external communication and provides an external IP once it's allocated by the external load balancer. It can handle any TCP connections or UDP flows, so it works for our use case.

This PR implements support for `LoadBalancer` and `NodePort` service types. Configured in `externalAccess.type`. The operator creates one external service per Segment Store instance.

In addition, these services are created with the `externalTrafficPolicy` option enabled. By default, `LoadBalancer` services route traffic to any backend pod in the cluster, however, with this option, services route traffic only to node-local pods.

#### Create `ClusterRole` permissions
When external access is enabled and configured with `NodePort`, the pod needs to find the node IP using the Kubernetes API. To achieve this, the service account needs to have `ClusterNode` read permissions on the `node` resource.

#### Segmentstore instance to advertise external address
By default, a Segmentstore instance uses it's local address when it registers with the Controller. However, the local pod address is only reachable from within the cluster (or namespace, depending on the environment). 

We need to initialize each Segmentstore instance with the external address allocated by the corresponding `LoadBalancer` service. To achieve this, the Pravega Docker image `entrypoint.sh` needs to modified to obtain the external service IP and port and configure it as system properties before starting the segmentstore process. 

The Segmentstore container needs to know the service name and the namespace to introspect and find its external IP. By naming convention, the `LoadBalancing` service name is the same as the pod name. So this is information can be easily made available to the pod by leveraging the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/). Two environment variables (`POD_NAME` and `POD_NAMESPACE`) are passed to the pod to allow the pods to find their information via the Kubernetes API. How the container uses this information is being addressed in the Pravega project PR https://github.com/pravega/pravega/pull/3062.

#### Ability to configure external access
Exposing the Pravega cluster to the outside is a possibility for users, but it's not enforced. For security reasons, by default a Pravega cluster will still be accessible only within the Kubernetes cluster. To enable external access, users need to explicitly create the `externalAccess` structure and set the `enabled` field to `true`, otherwise the cluster will not be exposed. E.g.:

```
apiVersion: "pravega.pravega.io/v1alpha1"
kind: "PravegaCluster"
metadata:
  name: "pravega1"
spec:
  zookeeperUri: 10.59.240.7:2181
  externalAccess:
    enabled: true
    type: LoadBalancer
...
```

## Purpose of the change
- Fix #61

## How to test the change

Assuming you already have access to a Kubernetes environment (GKE, PKS, Minikube...), follow the instructions in the README file to deploy ZooKeeper, the NFS server provisioner, and the Pravega Tier 2 PVC. Skip the deployment of the Pravega operator.

Check out this branch.
```
git checkout issue-61-external-traffic
```

Update the Pravega operator image to use an image built with the PR changes. Open the `deploy/operator.yaml` file and replace the container image from `pravega/pravega-operator:latest` to `adrianmo/pravega-operator:v0.1.0-23-dirty`. 

Deploy the operator.
```
kubectl create -f deploy/operator.yaml
```
Create a `pravega.yaml` that will use the updated images that contain the changes made to obtain and advertise the external service IP (ref: https://github.com/pravega/pravega/pull/3062) and the new `externalAccess` field.

```
apiVersion: "pravega.pravega.io/v1alpha1"                                                                                                                       
kind: "PravegaCluster"                                                                                                                                          
metadata:                                                                                                                                                       
  name: "pravega"                                                                                                                                               
spec:                                                                                                                                                           
  zookeeperUri: zk-client:2181                                                                                                                                  

  externalAccess:
    enabled: true
    type: LoadBalancer
                                                                                                                                                                
  bookkeeper:                                                                                                                                                                                                                                                                        
    image:
      repository: adrianmo/bookkeeper
      tag: 0.5.0-2023.5353487-SNAPSHOT
      pullPolicy: IfNotPresent

    replicas: 3

    storage:
      ledgerVolumeClaimTemplate:
        accessModes: [ "ReadWriteOnce" ]
        storageClassName: "standard"
        resources:
          requests:
            storage: 10Gi

      journalVolumeClaimTemplate:
        accessModes: [ "ReadWriteOnce" ]
        storageClassName: "standard"
        resources:
          requests:
            storage: 10Gi

    autoRecovery: true

  pravega:
    controllerReplicas: 1
    segmentStoreReplicas: 3

    cacheVolumeClaimTemplate:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "standard"
      resources:
        requests:
          storage: 20Gi

    image:
      repository: adrianmo/pravega
      tag: 0.5.0-2023.5353487-SNAPSHOT
      pullPolicy: IfNotPresent

    tier2:
      filesystem:
        persistentVolumeClaim:
          claimName: pravega-tier2
```

Copy the contents above to a `pravega.yaml` file and deploy the Pravega cluster:

```
kubectl create -f pravega.yaml
```
Watch the Pravega cluster being created...
```
watch kubectl get all -l pravega_cluster=pravega
```

Until the Controller and all Segmentstore services have been allocated an external address in the `EXTERNAL-IP` column. 

```
NAME                                              READY     STATUS    RESTARTS   AGE
pod/pravega-bookie-0                              1/1       Running   0          4m
pod/pravega-bookie-1                              1/1       Running   0          4m
pod/pravega-bookie-2                              1/1       Running   0          4m
pod/pravega-pravega-controller-76f6d74c55-kkwgb   1/1       Running   0          4m
pod/pravega-pravega-segmentstore-0                1/1       Running   0          4m
pod/pravega-pravega-segmentstore-1                1/1       Running   0          4m
pod/pravega-pravega-segmentstore-2                1/1       Running   0          4m

NAME                                            TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)                          AGE
service/pravega-bookie-headless                 ClusterIP      None            <none>           3181/TCP                         4m
service/pravega-pravega-controller              LoadBalancer   10.59.243.23    35.240.18.191    10080:31356/TCP,9090:32287/TCP   4m
service/pravega-pravega-segmentstore-0          LoadBalancer   10.59.248.137   35.241.234.206   12345:31849/TCP                  4m
service/pravega-pravega-segmentstore-1          LoadBalancer   10.59.243.10    35.189.250.112   12345:30177/TCP                  4m
service/pravega-pravega-segmentstore-2          LoadBalancer   10.59.252.97    35.240.48.116    12345:32460/TCP                  4m
service/pravega-pravega-segmentstore-headless   ClusterIP      None            <none>           12345/TCP                        4m

NAME                                               DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deployment.extensions/pravega-pravega-controller   1         1         1            1           4m

NAME                                                          DESIRED   CURRENT   READY     AGE
replicaset.extensions/pravega-pravega-controller-76f6d74c55   1         1         1         4m

NAME                                         DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/pravega-pravega-controller   1         1         1            1           4m

NAME                                                    DESIRED   CURRENT   READY     AGE
replicaset.apps/pravega-pravega-controller-76f6d74c55   1         1         1         4m

NAME                                            DESIRED   CURRENT   AGE
statefulset.apps/pravega-bookie                 3         3         4m
statefulset.apps/pravega-pravega-segmentstore   3         3         4m
```

Then, copy the Controller External IP and connect to it with the Pravega client. You can use the samples in the [Pravega samples repositories](https://github.com/pravega/pravega-samples).

---
Signed-off-by: Adrian Moreno <adrian@morenomartinez.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 
 go:
 - '1.10'
+- '1.11'
 
 sudo: required
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-- '1.11'
+- '1.10'
 
 sudo: required
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-- '1.10'
 - '1.11'
 
 sudo: required
--- a/pkg/apis/pravega/v1alpha1/types.go
+++ b/pkg/apis/pravega/v1alpha1/types.go
@@ -33,9 +33,10 @@ type PravegaCluster struct {
 }
 
 type PravegaClusterSpec struct {
-	ZookeeperUri string         `json:"zookeeperUri"`
-	Bookkeeper   BookkeeperSpec `json:"bookkeeper"`
-	Pravega      PravegaSpec    `json:"pravega"`
+	ZookeeperUri   string         `json:"zookeeperUri"`
+	ExternalAccess bool           `json:"externalAccess"`
+	Bookkeeper     BookkeeperSpec `json:"bookkeeper"`
+	Pravega        PravegaSpec    `json:"pravega"`
 }
 
 type ImageSpec struct {
--- a/pkg/apis/pravega/v1alpha1/types.go
+++ b/pkg/apis/pravega/v1alpha1/types.go
@@ -34,9 +34,14 @@ type PravegaCluster struct {
 
 type PravegaClusterSpec struct {
 	ZookeeperUri   string         `json:"zookeeperUri"`
-	ExternalAccess bool           `json:"externalAccess"`
+	ExternalAccess ExternalAccess `json:"externalAccess"`
 	Bookkeeper     BookkeeperSpec `json:"bookkeeper"`
 	Pravega        PravegaSpec    `json:"pravega"`
+}
+
+type ExternalAccess struct {
+	Enabled bool           `json:"enabled"`
+	Type    v1.ServiceType `json:"type,omitempty"`
 }
 
 type ImageSpec struct {
--- a/pkg/pravega/bookie.go
+++ b/pkg/pravega/bookie.go
@@ -17,7 +17,7 @@ import (
 	"github.com/pravega/pravega-operator/pkg/apis/pravega/v1alpha1"
 	"github.com/pravega/pravega-operator/pkg/utils/k8sutil"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/apps/v1beta1"
+	"k8s.io/api/apps/v1beta2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -84,11 +84,11 @@ func makeBookieHeadlessService(pravegaCluster *v1alpha1.PravegaCluster) *corev1.
 	}
 }
 
-func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta1.StatefulSet {
-	return &v1beta1.StatefulSet{
+func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta2.StatefulSet {
+	return &v1beta2.StatefulSet{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "StatefulSet",
-			APIVersion: "apps/v1beta1",
+			APIVersion: "apps/v1beta2",
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      k8sutil.StatefulSetNameForBookie(pravegaCluster.Name),
@@ -98,11 +98,14 @@ func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta1.Sta
 				*k8sutil.AsOwnerRef(pravegaCluster),
 			},
 		},
-		Spec: v1beta1.StatefulSetSpec{
-			ServiceName:          k8sutil.HeadlessServiceNameForBookie(pravegaCluster.Name),
-			Replicas:             &pravegaCluster.Spec.Bookkeeper.Replicas,
-			PodManagementPolicy:  v1beta1.ParallelPodManagement,
-			Template:             makeBookieStatefulTemplate(pravegaCluster),
+		Spec: v1beta2.StatefulSetSpec{
+			ServiceName:         k8sutil.HeadlessServiceNameForBookie(pravegaCluster.Name),
+			Replicas:            &pravegaCluster.Spec.Bookkeeper.Replicas,
+			PodManagementPolicy: v1beta2.ParallelPodManagement,
+			Template:            makeBookieStatefulTemplate(pravegaCluster),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: k8sutil.LabelsForBookie(pravegaCluster),
+			},
 			VolumeClaimTemplates: makeBookieVolumeClaimTemplates(&pravegaCluster.Spec.Bookkeeper),
 		},
 	}
--- a/pkg/pravega/bookie.go
+++ b/pkg/pravega/bookie.go
@@ -205,6 +205,7 @@ func makeBookieVolumeClaimTemplates(spec *v1alpha1.BookkeeperSpec) []corev1.Pers
 
 func makeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.ConfigMap {
 	configData := map[string]string{
+		"PRAVEGA_OPERATOR":         "true",
 		"BK_BOOKIE_EXTRA_OPTS":     "-Xms1g -Xmx1g -XX:MaxDirectMemorySize=1g -XX:+UseG1GC  -XX:MaxGCPauseMillis=10 -XX:+ParallelRefProcEnabled -XX:+UnlockExperimentalVMOptions -XX:+AggressiveOpts -XX:+DoEscapeAnalysis -XX:ParallelGCThreads=32 -XX:ConcGCThreads=32 -XX:G1NewSizePercent=50 -XX:+DisableExplicitGC -XX:-ResizePLAB",
 		"ZK_URL":                   pravegaCluster.Spec.ZookeeperUri,
 		"BK_useHostNameAsBookieID": "true",
--- a/pkg/pravega/bookie.go
+++ b/pkg/pravega/bookie.go
@@ -160,29 +160,6 @@ func makeBookiePodSpec(clusterName string, bookkeeperSpec *v1alpha1.BookkeeperSp
 				},
 			},
 		},
-		Affinity: &corev1.Affinity{
-			PodAntiAffinity: &corev1.PodAntiAffinity{
-				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
-					{
-						LabelSelector: &metav1.LabelSelector{
-							MatchExpressions: []metav1.LabelSelectorRequirement{
-								{
-									Key:      "component",
-									Operator: metav1.LabelSelectorOpIn,
-									Values:   []string{"bookie"},
-								},
-								{
-									Key:      "pravega_cluster",
-									Operator: metav1.LabelSelectorOpIn,
-									Values:   []string{clusterName},
-								},
-							},
-						},
-						TopologyKey: "kubernetes.io/hostname",
-					},
-				},
-			},
-		},
 	}
 }
 
--- a/pkg/pravega/bookie.go
+++ b/pkg/pravega/bookie.go
@@ -205,7 +205,6 @@ func makeBookieVolumeClaimTemplates(spec *v1alpha1.BookkeeperSpec) []corev1.Pers
 
 func makeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.ConfigMap {
 	configData := map[string]string{
-		"PRAVEGA_OPERATOR":         "true",
 		"BK_BOOKIE_EXTRA_OPTS":     "-Xms1g -Xmx1g -XX:MaxDirectMemorySize=1g -XX:+UseG1GC  -XX:MaxGCPauseMillis=10 -XX:+ParallelRefProcEnabled -XX:+UnlockExperimentalVMOptions -XX:+AggressiveOpts -XX:+DoEscapeAnalysis -XX:ParallelGCThreads=32 -XX:ConcGCThreads=32 -XX:G1NewSizePercent=50 -XX:+DisableExplicitGC -XX:-ResizePLAB",
 		"ZK_URL":                   pravegaCluster.Spec.ZookeeperUri,
 		"BK_useHostNameAsBookieID": "true",
--- a/pkg/pravega/bookie.go
+++ b/pkg/pravega/bookie.go
@@ -17,7 +17,7 @@ import (
 	"github.com/pravega/pravega-operator/pkg/apis/pravega/v1alpha1"
 	"github.com/pravega/pravega-operator/pkg/utils/k8sutil"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/apps/v1beta2"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -84,11 +84,11 @@ func makeBookieHeadlessService(pravegaCluster *v1alpha1.PravegaCluster) *corev1.
 	}
 }
 
-func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta2.StatefulSet {
-	return &v1beta2.StatefulSet{
+func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *appsv1.StatefulSet {
+	return &appsv1.StatefulSet{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "StatefulSet",
-			APIVersion: "apps/v1beta2",
+			APIVersion: "apps/v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      k8sutil.StatefulSetNameForBookie(pravegaCluster.Name),
@@ -98,10 +98,10 @@ func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta2.Sta
 				*k8sutil.AsOwnerRef(pravegaCluster),
 			},
 		},
-		Spec: v1beta2.StatefulSetSpec{
+		Spec: appsv1.StatefulSetSpec{
 			ServiceName:         k8sutil.HeadlessServiceNameForBookie(pravegaCluster.Name),
 			Replicas:            &pravegaCluster.Spec.Bookkeeper.Replicas,
-			PodManagementPolicy: v1beta2.ParallelPodManagement,
+			PodManagementPolicy: appsv1.ParallelPodManagement,
 			Template:            makeBookieStatefulTemplate(pravegaCluster),
 			Selector: &metav1.LabelSelector{
 				MatchLabels: k8sutil.LabelsForBookie(pravegaCluster),
--- a/pkg/pravega/bookie.go
+++ b/pkg/pravega/bookie.go
@@ -160,6 +160,29 @@ func makeBookiePodSpec(clusterName string, bookkeeperSpec *v1alpha1.BookkeeperSp
 				},
 			},
 		},
+		Affinity: &corev1.Affinity{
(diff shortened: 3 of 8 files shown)


pkg/pravega/pravega_segmentstore.go
```diff
@@ -245,3 +282,40 @@ func configureTier2Filesystem(podSpec *corev1.PodSpec, pravegaSpec *api.PravegaS
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      k8sutil.ServiceNameForSegmentStore(pravegaCluster.Name, i),
+				Namespace: pravegaCluster.Namespace,
+				Labels:    k8sutil.LabelsForSegmentStore(pravegaCluster),
+				OwnerReferences: []metav1.OwnerReference{
+					*k8sutil.AsOwnerRef(pravegaCluster),
+				},
+			},
+			Spec: corev1.ServiceSpec{
+				Type:                  corev1.ServiceTypeNodePort,
```
maddisondavid: Is there any way of telling the Operator to work WITHOUT a node port?  How does the Segmentstore know what IP address to register with the controller?

adrianmo: > Is there any way of telling the Operator to work WITHOUT a node port? 
Not yet. But it would be interesting to add the ability to enable/disable external connectivity to avoid having to create those extra resources.

> How does the Segmentstore know what IP address to register with the controller?
I'm still working on that part. The host IP can be obtained through the Downward API, but not sure yet about the node port exposed for that segmentstore instance...

pkg/pravega/bookie.go
```diff
@@ -84,11 +84,11 @@ func makeBookieHeadlessService(pravegaCluster *v1alpha1.PravegaCluster) *corev1.
 	}
 }
 
-func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta1.StatefulSet {
-	return &v1beta1.StatefulSet{
+func makeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *v1beta2.StatefulSet {
+	return &v1beta2.StatefulSet{
```
EronWright: I think that `apps/v1beta2/StatefulSet` is deprecated in favor of  `apps/v1/StatefulSet`.

pkg/apis/pravega/v1alpha1/types.go
```diff
@@ -33,9 +33,10 @@ type PravegaCluster struct {
 }
 
 type PravegaClusterSpec struct {
-	ZookeeperUri string         `json:"zookeeperUri"`
-	Bookkeeper   BookkeeperSpec `json:"bookkeeper"`
-	Pravega      PravegaSpec    `json:"pravega"`
+	ZookeeperUri   string         `json:"zookeeperUri"`
+	ExternalAccess bool           `json:"externalAccess"`
```
EronWright: I would strongly suggest that `externalAccess` be modeled as a struct, maybe with `enabled` field.   Over time there will be more parameters related to this.

pkg/pravega/pravega_controller.go
```diff
@@ -146,6 +149,12 @@ func makeControllerConfigMap(pravegaCluster *api.PravegaCluster) *corev1.ConfigM
 }
 
 func makeControllerService(pravegaCluster *api.PravegaCluster) *corev1.Service {
+
+	serviceType := corev1.ServiceTypeClusterIP
+	if pravegaCluster.Spec.ExternalAccess {
+		serviceType = corev1.ServiceTypeLoadBalancer
```
EronWright: Given that the use of an LB here is a workaround, as opposed to a NodePort, we should make it explicit so that we can use NodePort as the default approach in future.  e.g.
```
externalAccess:
  enabled: true
  type: LoadBalancer
```

pkg/pravega/pravega_segmentstore.go
```diff
@@ -116,13 +138,66 @@ func makeSegmentstorePodSpec(pravegaCluster *api.PravegaCluster) corev1.PodSpec
...
+						TopologyKey: "kubernetes.io/hostname",
+					},
+				},
+			},
+		},
+		Volumes: []corev1.Volume{
+			{
+				Name: "podinfo",
+				VolumeSource: corev1.VolumeSource{
+					DownwardAPI: &corev1.DownwardAPIVolumeSource{
```
EronWright: An surprised that you exposed the downward API as a file instead of as a set of environment variables.  Rationale?

pkg/pravega/pravega_segmentstore.go
```diff
@@ -245,3 +324,70 @@ func configureTier2Filesystem(podSpec *corev1.PodSpec, pravegaSpec *api.PravegaS
...
+			Labels:    k8sutil.LabelsForSegmentStore(pravegaCluster),
+			OwnerReferences: []metav1.OwnerReference{
+				*k8sutil.AsOwnerRef(pravegaCluster),
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Name:     "server",
+					Port:     12345,
```
EronWright: Where does 12345 come from?  Seems to me that the normal segment store port should be used here.   Meanwhile, in a NodePort situation, K8s supplies the port AFAIK.

pkg/utils/k8sutil/pravega.go
```diff
@@ -17,21 +17,29 @@ import (
 )
 
 func ConfigMapNameForBookie(clusterName string) string {
-	return fmt.Sprintf("%s-bookie-config", clusterName)
```
EronWright: Please try not to bundle unrelated work.

pkg/pravega/pravega_segmentstore.go
```diff
@@ -34,6 +35,21 @@ const (
 )
 
 func deploySegmentStore(pravegaCluster *api.PravegaCluster) (err error) {
+	err = sdk.Create(makeSegmentStoreHeadlessService(pravegaCluster))
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+
+	if pravegaCluster.Spec.ExternalAccess {
+		services := makeSegmentStoreLoadBalancerServices(pravegaCluster)
```
EronWright: This logic will probably require revision once we support scaling of the segment stores.  Since you'd need to create/delete service resources as needed.

pkg/utils/k8sutil/pravega.go
```diff
@@ -41,11 +49,11 @@ func DeploymentNameForController(clusterName string) string {
 }
 
 func ConfigMapNameForSegmentstore(clusterName string) string {
-	return fmt.Sprintf("%s-segmentstore", clusterName)
+	return fmt.Sprintf("%s-pravega-segmentstore", clusterName)
```
EronWright: Given that the cluster name is the prefix, and will have the value `pravega` in most cases, it seems flat wrong to add `-pravega-`. What does it accomplish?

EronWright:
It would be great to see the output of `kubectl get -n nautilus-pravega all`, attached to this PR, thanks!

pkg/utils/k8sutil/pravega.go
adrianmo: The cluster name is user defined, so it might or might not contain "pravega". I've added the `- pravega` prefix to be consistent with other components that already use the same prefix and to be more verbose and indicate that it's a Pravega resource.

pkg/utils/k8sutil/pravega.go
adrianmo: You are right, I added this because it's a small naming change and I was just trying be consistent in the way we name resources. It doesn't make sense to add suffixes like `-config`, `-service`, `-deployment` because they are already differentiated by their resource type.

pkg/pravega/pravega_segmentstore.go
adrianmo: Isn't `12345` the default port for the Segment Store?

adrianmo: https://github.com/pravega/pravega/blob/master/config/config.properties#L55

pkg/apis/pravega/v1alpha1/types.go
adrianmo: I like the idea of making external access configurable. I'll create a struct as you suggested.

adrianmo:
@EronWright thanks for the review. I'll make some changes based on your comments.

pkg/pravega/pravega_controller.go
EronWright: Nonetheless I strongly feel it should be an explicit option, not the default.

pkg/pravega/pravega_segmentstore.go
EronWright: Hmm did not know that.   Wow does that suck!

pkg/pravega/bookie.go
adrianmo: Updated. Thanks for the heads up.

pkg/apis/pravega/v1alpha1/types.go
adrianmo: Updated. I've created a struct for the external access configuration.

pkg/pravega/pravega_controller.go
adrianmo: I've added support for `NodePort` service type. However, I've set `LoadBalancer` as the default option because it's the only option that works in GKE and PKS at this moment.

adrianmo:
I've made a few changes suggested by @EronWright and updated the PR description.

pkg/pravega/pravega_segmentstore.go
adrianmo: Updated. Downward API is now exposed via env vars.

adrianmo:
@EronWright @maddisondavid can you please review the PR again after the changes?

pkg/apis/pravega/v1alpha1/types.go
```diff
@@ -33,9 +33,15 @@ type PravegaCluster struct {
...
-	Bookkeeper   BookkeeperSpec `json:"bookkeeper"`
-	Pravega      PravegaSpec    `json:"pravega"`
+	ZookeeperUri   string         `json:"zookeeperUri"`
+	ExternalAccess ExternalAccess `json:"externalAccess"`
+	Bookkeeper     BookkeeperSpec `json:"bookkeeper"`
+	Pravega        PravegaSpec    `json:"pravega"`
+}
+
+type ExternalAccess struct {
+	Enabled bool           `json:"enabled"`
```
fpj: My understanding is that the default should be false (for security). Do we set such a default anywhere or is it implied by the absence of a value?

pkg/pravega/pravega_segmentstore.go
```diff
@@ -34,6 +35,21 @@ const (
 )
 
 func deploySegmentStore(pravegaCluster *api.PravegaCluster) (err error) {
+	err = sdk.Create(makeSegmentStoreHeadlessService(pravegaCluster))
```
fpj: Is this necessary even in the case that external access is disabled?

pkg/pravega/pravega_segmentstore.go
```diff
@@ -123,6 +146,29 @@ func makeSegmentstorePodSpec(pravegaCluster *api.PravegaCluster) corev1.PodSpec
 				},
 			},
 		},
+		Affinity: &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
```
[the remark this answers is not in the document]
adrianmo: Yes, I implemented affinity rules both in this PR and in #92. The ones in #92 are more complete and also include rules for Controller instances, so I've removed the affinity rules in this PR.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-- '1.10'
+- '1.11'
```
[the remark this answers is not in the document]
adrianmo: Yes, this is unrelated to the issue. I've changed the Go version back to 1.10.

pkg/pravega/pravega_segmentstore.go
fpj: Please bear with me, I'm almost there. A few points that called my attention:

- This call to make each segment store a headless service wasn't there before, yet we were able to get Pravega up and running. If that's the case, then d we still need to make this call in the absence of external access?
- I'm not sure I understand when you say that segment store instances communicate with other. I think you mean to say controller to segment store communication, as segment store instances do not talk to each other as of today.

adrianmo: > This call to make each segment store a headless service wasn't there before, yet we were able to get Pravega up and running. If that's the case, then do we still need to make this call in the absence of external access?

You are right. This is actually not necessary for external connectivity to work, but it was added there as a best practice for internal communication when creating stateful apps in K8.

> I'm not sure I understand when you say that segment store instances communicate with each other. I think you mean to say controller to segment store communication, as segment store instances do not talk to each other as of today.

Correct. Controller communication with Segmentstores happens via external IP address when external connectivity is enabled. Even though this works, it's not ideal as we are adding unnecessary hops to the traffic route. Ideally, the Controller would use the headless service to communicate with individual Segmentstores. Before we can get this implemented in the operator, Pravega needs to support different endpoints for inter-component communication and external clients.

PR #89, opened 2018-11-19
Issue 88: Set default values when not specified

adrianmo:
## Change log description
- If a `PravegaCluster` manifest is submitted with partially filled (or even with an empty spec), the operator will complete the resource specification with the default values outlined in #88. 
- Fix `PravegaCluster` spec not showing when describing a resource (e.g. `kubectl describe PravegaCluster pravega`).
- Add descriptive comments to API fields

## Purpose of the change
- Fix #88 

## How to test the change

Checkout the repository (`master` branch is fine, no need to checkout this branch).

Edit the operator image used in the `deploy/operator.yaml` file to use the `adrianmo/pravega-operator:issue-88` image, which contains the changes in this PR.

Follow the regular README file to install the dependencies and the operator.

Create and submit a CRD manifest with an empty spec.

```
apiVersion: "pravega.pravega.io/v1alpha1"
kind: "PravegaCluster"
metadata:
  name: "pravega"
```

Check that resources are created as usual.

```
$ kubectl get all -l pravega_cluster=pravega
NAME                                              READY     STATUS    RESTARTS   AGE                                                                           
pod/pravega-bookie-0                              1/1       Running   0          16m                                                                           
pod/pravega-bookie-1                              1/1       Running   0          16m                                                                           
pod/pravega-bookie-2                              1/1       Running   0          16m                                                                           
pod/pravega-pravega-controller-84f6b77c64-gjvlt   1/1       Running   0          16m                                                                           
pod/pravega-pravega-segmentstore-0                1/1       Running   1          16m                                                                           

NAME                                            TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)              AGE                                             
service/pravega-bookie-headless                 ClusterIP   None            <none>        3181/TCP             16m                                             
service/pravega-pravega-controller              ClusterIP   10.59.251.157   <none>        10080/TCP,9090/TCP   16m                                             
service/pravega-pravega-segmentstore-headless   ClusterIP   None            <none>        12345/TCP            16m                                             

NAME                                               DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE                                                            
deployment.extensions/pravega-pravega-controller   1         1         1            1           16m                                                            

NAME                                                          DESIRED   CURRENT   READY     AGE                                                                
replicaset.extensions/pravega-pravega-controller-84f6b77c64   1         1         1         16m                                                                

NAME                                         DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE                                                                  
[...]

Issue #7, opened 2018-07-12
Admission Controller for PravegaCluster resource


maddisondavid:
Currently the resource contains the Bookkeeper and Pravega image definitions.  The operator should have sensible defaults set in its configuration and apply these to the resource via a mutating Admission Controller.

adrianmo:
Blocked by: https://github.com/operator-framework/operator-sdk/issues/269

adrianmo:
Related to #88

adrianmo:
The admission controller should also attempt to mutate the values in the `options` block.

https://github.com/pravega/pravega-operator/blob/78d99444549be1060e6883e9dea13c75b74c9fca/pkg/apis/pravega/v1alpha1/pravega.go#L95

Option values are expected to be strings, surrounded by double quotes, otherwise the operator framework will fail to parse it and reject the CR. It'd be interesting if the admission controller could transform to strings any values passed as options.

PR #92, opened 2018-11-20
Issue 60: Data protection / High availability

### LINKED ISSUE #60: Ensure data protection (DU/DL)
Operator should ensure that components (e.g. bookies) are deployed in a reasonable way from a data protection perspective.   Kubernetes has features including [anti-affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity-beta-feature), multi-zone, and [pod disruption budgets](https://kubernetes.io/docs/tasks/run-application/configure-pdb/).

Achieving adequate data protection likely has deployment- and upgrade-time considerations, e.g. anti-affinity to prevent DL, pod disruption budget to prevent DU during upgrade/maintenance.  Note that Kubernetes has support for "planned" node maintenance operations, e.g. `kubectl drain` ([ref](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/)).  Such operations respect the pod disruption budget.

For a conceptual overview, please read the [Disruptions](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/) section of the Kubernetes documentation.

adrianmo:
## Change log description
The operator should ensure that pods are deployed in a reasonable way from a data protection and high availability perspective.

#### Pod protection
Kubernetes has the ability to safely [evict all of your pods from a node](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/) before an admin can perform maintenance on a node. The operator should prevent this from happening if the node evicted would cause disruption in a Pravega cluster. This can be achieved with [Pod Disruption Budgets](https://kubernetes.io/docs/tasks/run-application/configure-pdb/).

With these changes, the operator will create a PDB for each component type (controller, segmentstore, bookie) with the following PDB configuration.

- Controller: allow planned disruptions as long as there is at one Controller pod available.
- Segment Store: allow 1 planned disruption if the number of replicas is greater than 1, otherwise, no planned disruption is allowed.
- Bookkeeper: allow 1 planned disruption at a time. 
  - The minimum number of Bookies is enforced to 3. 
  - The quorum has been brought down to 2 ([ref](https://github.com/pravega/pravega/pull/3158)) This allows one Bookie disruption even in the minimum configuration.
  - `AutoRecovery` is set to `true` by default to automatically recover under-replicated ledger when there is a pod disruption (planned or unplanned).

#### Pod placement

To achieve higher performance and fault tolerance instances of the same type are recommended to be spread across different nodes. For example, assuming that we have a 3-node Kubernetes cluster and we are attempting to create 3 Bookkeeper instances, the first Bookie will be placed in any of the 3 nodes (e.g. _Node 1_), the second one will be preferably placed in a different node (e.g. _Node 2_ or _Node 3_), and the third bookie will be placed in the only node that does not contain a bookie. If a fourth Bookie was to be scheduled, it would be placed in any of the nodes, because all of them already have one Bookie. 

This has been implemented in this PR with [Pod Anti-Affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity-beta-feature) using the `preferredDuringSchedulingIgnoredDuringExecution` soft requirement because there's no functional requirement to place components of the same kind in different nodes.

## Purpose of the change
- Fix #60 

## How to test the change
Assuming you already have access to a Kubernetes environment (GKE, PKS...) with at least 3 nodes, follow the instructions in the README file to deploy ZooKeeper, the NFS server provisioner, and the Pravega Tier 2 PVC. Skip the deployment of the Pravega operator.

Check out this branch.
```
git checkout issue-60-data-protection-ha
```

Update the Pravega operator image to use an image built with the PR changes. Open the `deploy/operator.yaml` file and replace the container image from `pravega/pravega-operator:latest` to `adrianmo/pravega-operator:issue-60`. 

Deploy the operator.
```
kubectl create -f deploy/operator.yaml
```
Follow the instructions in the README file to deploy a Pravega cluster with 3 Bookies, 1 or more Controllers, and 3 Segment stores.

Run the following command to verify the correct placement of pods.

```
$ kubectl get po -o wide
NAME                                         READY     STATUS    RESTARTS   AGE       IP           NODE
...
pravega-bookie-0                             1/1       Running   0          35s       10.56.6.43   gke-adrian-cluster-1-default-pool-6e9f73a5-7348
pravega-bookie-1                             1/1       Running   0          35s       10.56.1.29   gke-adrian-cluster-1-default-pool-0616952c-kd4z
pravega-bookie-2                             1/1       Running   0          35s       10.56.9.42   gke-adrian-cluster-1-default-pool-d0ef8647-f01m
pravega-operator-7dbd5bb757-wzp7z            1/1       Running   0          1h        10.56.9.38   gke-adrian-cluster-1-default-pool-d0ef8647-f01m
pravega-pravega-controller-58b4dcf4f-69jpv   1/1       Running   0          35s       10.56.1.28   gke-adrian-cluster-1-default-pool-0616952c-kd4z
pravega-segmentstore-0                       1/1       Running   0          35s       10.56.3.21   gke-adrian-cluster-1-default-pool-0616952c-n948
pravega-segmentstore-1                       1/1       Running   0          35s       10.56.9.41   gke-adrian-cluster-1-default-pool-d0ef8647-f01m
pravega-segmentstore-2                       1/1       Running   0          35s       10.56.6.42   gke-adrian-cluster-1-default-pool-6e9f73a5-7348
...
```
Verify that pods of the same kind are scheduled to different nodes.

Check that pod disruption budgets are correctly set.

```
$ kubectl get pdb
NAME                         MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
pravega-bookie               N/A             1                 1                     6m
pravega-pravega-controller   1               N/A               0                     6m
pravega-segmentstore         N/A             1                 1                     6m
```
In this example, we allow one pod disruption for Segment Store and Bookkeeper, but none disruption for the Controller since there's only one Controller pod.

If we try to drain the node that contains the Controller instance, we should expect the drain command to abort.

```
$ kubectl drain gke-adrian-cluster-1-default-pool-0616 --ignore-daemonsets
952c-kd4z                                                                                                                                                       
node "gke-adrian-cluster-1-default-pool-0616952c-kd4z" cordoned                                                                                                 
error: unable to drain node "gke-adrian-cluster-1-default-pool-0616952c-kd4z", aborting command...
```

```
$ kubectl get nodes
NAME                                              STATUS                     ROLES     AGE       VERSION
...
gke-adrian-cluster-1-default-pool-0616952c-kd4z   Ready,SchedulingDisabled   <none>    4d        v1.9.7-gke.11
...
```
The node will continue to be part of the cluster, but it will not host any new pods. The controller instance will still be untouched and the Pravega admin will be required to manually increase the number of replicas and/or delete the affected pod to automatically be rescheduled to a different node.


adrianmo:
[Data Protection and High Availability in K8](https://trello.com/c/z7bDYbKn/31-data-protection-and-high-availability-in-k8)

pkg/pravega/pravega_controller.go
```diff
@@ -180,3 +214,37 @@ func makeControllerService(pravegaCluster *api.PravegaCluster) *corev1.Service {
 		},
 	}
 }
+
+func makeControllerPodDisruptionBudget(pravegaCluster *api.PravegaCluster) *policyv1beta1.PodDisruptionBudget {
+	var maxUnavailable intstr.IntOrString
+
+	if pravegaCluster.Spec.Pravega.ControllerReplicas == int32(1) {
```
fpj: Can't we express this as min available of 1 rather than having two options for max unavailable? it is just to simplify the logic.

pkg/pravega/pravega_segmentstore.go
```diff
@@ -188,6 +222,40 @@ func makeCacheVolumeClaimTemplate(pravegaSpec *api.PravegaSpec) []corev1.Persist
 	}
 }
 
+func makeSegmentstorePodDisruptionBudget(pravegaCluster *api.PravegaCluster) *policyv1beta1.PodDisruptionBudget {
+	var maxUnavailable intstr.IntOrString
+
+	if pravegaCluster.Spec.Pravega.SegmentStoreReplicas == int32(1) {
```
fpj: Same comment about min available and max unavailable.

pkg/pravega/bookie.go
```diff
@@ -157,6 +165,32 @@ func makeBookiePodSpec(clusterName string, bookkeeperSpec *v1alpha1.BookkeeperSp
 				},
 			},
 		},
+		Affinity: &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+					{
+						Weight: 100,
```
fpj: What's the criteria for choosing 100 for the weight? I'm trying to understand why `100` and not `99` or `10`or something else.

pkg/pravega/pravega_controller.go
adrianmo: Well, the logic will be different if we change it to `minAvailable=1`.

- `maxUnavailable=1` means that, regardless of the number of Controller replicas, only one of them can be disrupted. E.g. if we have 5 replicas, we can only disrupt 1.
- `minAvailable=1` means that we allow disruptions as long as we have one Controller replica. E.g. if we have 5 replicas, we allow 4 disruptions.

In the existing, we allow 1 planned disruption if we have more than 1 controller, 0 otherwise. Does this behavior look good to you?

pkg/pravega/pravega_segmentstore.go
adrianmo: Please check response in controller comment.

pkg/pravega/bookie.go
adrianmo: As we only have one affinity rule, I set it to 100 (max priority) following the example in https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity

pkg/pravega/bookie.go
```diff
@@ -211,3 +245,31 @@ func makeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.Config
 		Data: configData,
 	}
 }
+
+func makeBookiePodDisruptionBudget(pravegaCluster *v1alpha1.PravegaCluster) *policyv1beta1.PodDisruptionBudget {
+	minAvailable := intstr.FromInt(3)
```
shrids: This implies if the bookie count is `x, where x > 3` then `x-3` bookies can go down at the same time, right?
Also, should the behavior change based on `auto_recovery:  true/false` ?

pkg/pravega/bookie.go
```diff
@@ -30,6 +33,11 @@ const (
 
 func deployBookie(pravegaCluster *v1alpha1.PravegaCluster) (err error) {
 
+	err = sdk.Create(makeBookiePodDisruptionBudget(pravegaCluster))
```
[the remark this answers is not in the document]
adrianmo: As stated in the README file, the operator requires Kubernetes version 1.8 or higher. Therefore, the draining feature will always be available.

pkg/pravega/bookie.go
adrianmo: > This implies if the bookie count is x, where x > 3 then x-3 bookies can go down at the same time, right?

Correct. There will always be at least 3 bookies (planned disruptions). Does it sound reasonable to you from a data protection standpoint?

> Also, should the behavior change based on auto_recovery: true/false ?

Good point, but I'm not sure if it makes sense to make the distinction... Do you have any insight or suggestion?

adrianmo:
Need to rebase and fix conflicts.

adrianmo:
Blocked on #89 that brings the ability to set default values.

adrianmo:
@EronWright the anti-affinity rules used in this PR are soft requirements (`PreferredDuringSchedulingIgnoredDuringExecution`), meaning that if there's only one node (e.g. MiniKube), all pods will run in that node.

adrianmo:
Need to bring down the BK replication factor to 2 as it was done in https://github.com/pravega/pravega/pull/3158

adrianmo:
@fpj @shrids The PR is ready to review again.

Recent changes:
- Enforce the minimum number of Bookies to 3 (inherited by #89).
- Set `AutoRecovery` to `true` by default (inherited by #89).
- Bring down Bookie quorum to 2 (as in https://github.com/pravega/pravega/pull/3158).
- Allow only one Bookie pod disruption at a time.

pkg/controller/pravega/pravega_controller.go
```diff
@@ -160,3 +163,30 @@ func MakeControllerService(p *api.PravegaCluster) *corev1.Service {
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "PodDisruptionBudget",
+			APIVersion: "policy/v1beta1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      util.PdbNameForController(pravegaCluster.Name),
+			Namespace: pravegaCluster.Namespace,
+		},
+		Spec: policyv1beta1.PodDisruptionBudgetSpec{
+			MaxUnavailable: &maxUnavailable,
```
shrids: Could you elaborate on the reason why we are not using MinUnavailable for controller? (if we have multiple controllers on the same node, we can drain all the controllers provided there are instances of a controller running other nodes...)

Specifying MinUnavailable = 1 and removing MaxUnavailable(as they are mutually exclusive settings) we could remove `if else ` at line 170.

adrianmo: You are right, being a stateless component, `minAvailable=1` makes more sense. Updated it.

adrianmo:
Updated PR and description to address @shrids comments. Changed Controller PDB to `minAvailable=1`.

fpj:
Are you planning on adding any sort of test cases here?

adrianmo:
Yes, I'll add them today.

test/e2e/e2eutil/pravegacluster_util.go
```diff
@@ -0,0 +1,104 @@
...
+// CreateCluster creates a PravegaCluster CR with the desired spec
+func CreateCluster(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, p *api.PravegaCluster) (*api.PravegaCluster, error) {
+	t.Logf("creating pravega cluster: %v", p.Name)
+	err := f.Client.Create(goctx.TODO(), p, &framework.CleanupOptions{TestContext: ctx, Timeout: CleanupTimeout, RetryInterval: CleanupRetryInterval})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create CR: %v", err)
+	}
+
+	// Give the operator a few seconds to fill the PravegaCluster spec with default values if necessary
+	time.Sleep(5 * time.Second)
```
Tristan1900: I am wondering what if the cluster is very slow? I see some examples are first using waitForPravegaCluster to make sure the cluster is running, then getting CR will be reliable.

adrianmo: This line is not waiting for the Pravega cluster to come up, it's just giving a few seconds for the operator to process the CR and fill the missing parameters with default values. 5 seconds should be more than enough for the operator to do it. However, I'm open to other, more reliable alternatives.

EronWright: We should be using a mutating webhook to apply defaults, so that they are applied synchronously.

fpj:
I wanted to capture an offline discussion with @EronWright . BK automatically replicates ledger fragments automatically (when the feature is enabled). Even if we have a disruption budget such that we have a single disruption at a time, there is no guarantee that the data will be re-replicated fast enough. As such, we could end up in this situation in which we get rid of too many copies of the data and can't recover it.

Ideally, we wait for the re-replication to finish before causing another disruption, but I don't think we have an API for that. If we are replacing bookies, then we could consider keeping the volumes of the decommissioned bookies, but that's confusing in the general case.

test/e2e/e2eutil/pravegacluster_util.go
adrianmo: @EronWright that will happen when we implement an [Admission Controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/) to validate CR objects and apply default values. This work is tracked here: https://github.com/pravega/pravega-operator/issues/7

adrianmo:
Related to @fpj's comment in https://github.com/pravega/pravega-operator/pull/92#issuecomment-447960089, I've created issue https://github.com/pravega/pravega-operator/issues/114 to investigate BookKeeper's disruption cases.

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -326,3 +331,30 @@ func MakeSegmentStoreExternalServices(pravegaCluster *api.PravegaCluster) []*cor
 	}
 	return services
 }
+
+func MakeSegmentstorePodDisruptionBudget(pravegaCluster *api.PravegaCluster) *policyv1beta1.PodDisruptionBudget {
+	var maxUnavailable intstr.IntOrString
+
+	if pravegaCluster.Spec.Pravega.SegmentStoreReplicas == int32(1) {
+		maxUnavailable = intstr.FromInt(0)
```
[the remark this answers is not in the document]
adrianmo: PDB protects pods from planned disruptions, such as a node eviction. In such case, if we have 1 Segment Store replica and `maxUnavailable=1`, Kubernetes will first kill the pod, and then start a new one in a different node, causing a temporary disruption. Whereas if `maxUnavailble=0`, Kubernetes will put node eviction on hold, giving the user time to increase the replica number and create a new pod in a different node, and then delete the affected pod to resume the node eviction, resulting in zero downtime.

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -141,6 +143,9 @@ func makeSegmentstorePodSpec(pravegaCluster *api.PravegaCluster) corev1.PodSpec
 func MakeSegmentstoreConfigMap(pravegaCluster *api.PravegaCluster) *corev1.ConfigMap {
 	javaOpts := []string{
 		"-Dpravegaservice.clusterName=" + pravegaCluster.Name,
+		"-Dbookkeeper.bkEnsembleSize=2",
```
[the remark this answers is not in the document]
adrianmo: Good catch. I'll remove those lines as the default value is already set to `2` and this way we will allow users to override the default values using the `options` CR section.

```
...
  pravega:
    controllerReplicas: 1
    segmentStoreReplicas: 3

    # See https://github.com/pravega/pravega/blob/3f5b65084ae17e74c8ef8e6a40e78e61fa98737b/config/config.properties
    # for available configuration properties
    options:
      bookkeeper.bkEnsembleSize=2
      bookkeeper.bkAckQuorumSize=2
      bookkeeper.bkWriteQuorumSize=2
...
```

pkg/controller/pravega/pravega_segmentstore.go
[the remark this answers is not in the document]
fpj: I don't understand this comment:

> if maxUnavailble=0, Kubernetes will put node eviction on hold, giving the user time to increase the replica number and create a new pod in a different node, and then delete the affected pod to resume the node eviction, resulting in zero downtime.

How does the user know it is being given time to increase the replica count? Also, is it saying that we can have multiple instances temporarily? That might be worse than having none temporarily and then one back because that will multiple temporarily will induce two rebalances of the segment containers.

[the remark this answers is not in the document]
adrianmo: > How does the user know it is being given time to increase the replica count? 

Due to the PDB, Kubernetes will put the eviction on hold until the affected pod is deleted. It's up to the user to decide what to do, they can increase the replica count to avoid down time; or they can just delete the pod altogether and let Kubernetes reschedule it.

> Also, is it saying that we can have multiple instances temporarily?

Yes, but not necessarily, the user can just delete the pod and avoid having multiple instances temporarily.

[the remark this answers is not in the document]
fpj: I'm confused about how the user knows that a pod needs to be deleted. I'm assuming that the eviction is part of some automated process and K8s is applying a policy to decide how to bring specific pods down.

PR #122, opened 2019-01-21
Issue 64: configure bookie metrics

### LINKED ISSUE #63: Configure Bookie metrics
The operator should support bookie metrics, notably graphite (as provided by codahale) and prometheus.

Related documentation:
http://bookkeeper.apache.org/docs/4.7.0/admin/metrics/

### LINKED ISSUE #116: BookKeeper client is not able recover from restarted bookie
The [BookieFailoverTest](https://github.com/pravega/pravega/blob/master/test/system/src/test/java/io/pravega/test/system/BookieFailoverTest.java) 
case (Pravega system tests) is failing consistently because the segment store is not able to recover from a bookie that is stopped and later restarted (there are exactly three bookies and the ledger ensemble requires three bookies).

The problem, observed from the logs, is that the BK client that the segment store instantiates detects that the bookie has crashed and it tries to get a new bookie. As part of the process of selecting a new bookie, it excludes bookies that it has used previously in the ledger. The client continues to attempt a connection to the previous bookie periodically, but the new incarnation happens to have a different IP address. The client uses the bookie host name, which keeps resolving to the old address and any connection to it fails.

Here is some relevant detail. In this line:

https://github.com/pravega/pravega-operator/blob/99901541159c81056f160a97274baeb6ea337ae6/pkg/controller/pravega/bookie.go#L180 

We ask the bookies to use the host name rather than the IP address. The client receives the host name as the client ID and uses the host name to connect.

This test case used to pass up to the Pravega `r0.3`. After `r0.3`, we introduced this change:

https://github.com/pravega/pravega/pull/2833

A solution here would be to make sure that the bookie host name is unique despite restarts.

Tristan1900:
### Change log description
- Enable bookkeeper metrics configuration.
- The user can decide whether to turn on bookkeeper metrics and to choose metrics provider by specifying it in Kubernetes manifest. 

### Purpose of the change
Fix #63 
Fix #116

### How to test the change
A possible way to test: choose a stats provider like Prometheus, deploy Prometheus in k8s and port forward it to local, then use a browser to see if there are any metrics. 
Another way: choose codahale and specify a [csv endpoint](https://github.com/apache/bookkeeper/blob/master/conf/bk_server.conf#L800), use a script to check if there are any files created.


pkg/controller/pravega/bookie.go
```diff
@@ -191,6 +191,10 @@ func MakeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.Config
 		configData["BK_AUTORECOVERY"] = "true"
 	}
 
+	for k, v := range pravegaCluster.Spec.Bookkeeper.Options {
+		configData[k] = v
```
adrianmo: @Tristan1900 is this working to you? I thought you had to add the `BK_` prefix to the options.

adrianmo:
Blocked on https://github.com/pravega/pravega/pull/3357

adrianmo:
Backwards compatibility changes made. Jenkins job in progress: https://asd-nautilus-jenkins.isus.emc.com/view/Pravega/job/pravega-branch-builder/913/

Tristan1900:
Tested against the latest Pravega master branch.
Used codahale provider and specified a csv endpoint, confirmed that csv file were written to that directroy when Pravega cluster is running.
```
options:
         enableStatistics: "true"
         statsProviderClass: "org.apache.bookkeeper.stats.codahale.CodahaleMetricsProvider"
         codahaleStatsCSVEndpoint: "/tmp"
```

The default provider Prometheus is also available. 
But be aware that the Twitter stats providers are still not accessible.

README.md
```diff
@@ -26,7 +26,7 @@ The project is currently alpha. While no breaking API changes are currently plan
 * [Releases](#releases)
 * [Troubleshooting](#troubleshooting)
     * [Helm Error: no available release name found](#helm-error-no-available-release-name-found)
-
+    * [NFS volume mount failure: wrong fs type](#nfs-volume-mount-failure-wrong-fs-type)
```
fpj: This change is unrelated to this PR.

README.md
```diff
@@ -501,3 +501,27 @@ kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admi
 kubectl patch deploy --namespace kube-system tiller-deploy -p '{"spec":{"template":{"spec":{"serviceAccount":"tiller"}}}}'
 ```
 The above commands should resolve the errors and `helm install` should work correctly.
+
+### NFS volume mount failure: wrong fs type
```
fpj: This change is also unrelated to this PR.

example/cr-detailed.yaml
```diff
@@ -32,6 +32,12 @@ spec:
     # see https://bookkeeper.apache.org/docs/latest/admin/autorecovery/
     autoRecovery: true
 
+    # To enable bookkeeper metrics feature
```
fpj: Unrelated to this PR, I noticed that this file does not have a license header.

example/cr-detailed.yaml
```diff
@@ -32,6 +32,12 @@ spec:
...
+    # See http://bookkeeper.apache.org/docs/4.7.0/admin/metrics/ for more metrics provider
+    options:
+      enableStatistics: "true"
+      statsProviderClass: "org.apache.bookkeeper.stats.codahale.CodahaleMetricsProvider"
```
fpj: Is this configuration sufficient? Don't we need to configure endpoints?

http://bookkeeper.apache.org/docs/4.7.0/reference/config/#statistics

README.md
adrianmo: That's right. I'll move this to a separate branch.

example/cr-detailed.yaml
adrianmo: This file is just a sample CR definition that can be taken and modified by anyone and it does not contain any intellectual property.

example/cr-detailed.yaml
adrianmo: Right, for this to work, user would need to configure the endpoint. @Tristan1900 can you add a sample endpoint configuration?

README.md
adrianmo: Done.

README.md
adrianmo: Removed from PR.

example/cr-detailed.yaml
kevinhan88: Yes, please call out the connection details (protocol, host and port) for CodahaleMetricsProvider and the default values.

kevinhan88:
Just a thought: with the micrometer migration, currently Pravega supports Statsd and InfluxDB but not CSV and Prometheus. We may need to keep things consistent across Pravega and Bookie in documents.
For example, if we mention CSV here but user may expect metrics from both bookie and Pravega; similarly for Prometheus.

example/cr-detailed.yaml
kevinhan88: Also please call out the output frequency. From operation perspective it may be necessary to have consistent output frequency across Bookie and Pravega, to easy the querying and reading of metrics.

Tristan1900: Thanks for the comments guys. I am adding more details to this.

Tristan1900:
Thanks for the heads up @kevinhan88.
I am using CSV just to verify if the provider works in this PR, I can exclude it from the `cr-detailed.yaml`.

Tristan1900:
@kevinhan88 Right. I don't think that format is supported. This [line](https://github.com/apache/bookkeeper/blob/master/bookkeeper-stats-providers/codahale-metrics-provider/src/main/java/org/apache/bookkeeper/stats/CodahaleMetricsProvider.java#L79) in BK repo only accepts `hostname:port` or `ip:port` format. [Here](https://metrics.dropwizard.io/3.1.0/manual/graphite/) is the official guide of how to use the client.
As for protocol, I found TCP is what the client is using. See [here](https://metrics.dropwizard.io/3.1.0/apidocs/com/codahale/metrics/graphite/Graphite.html)

PR #123, opened 2019-01-28
Issue 10: Update cluster status

### LINKED ISSUE #10: Support for `PravegaCluster` resource status
Report the high-level status / conditions via the status section of the custom resource.  A key scenario is to enable Helm to `--wait` on the deployment of the cluster.

Ideally the status would accurately indicate whether Pravega is ready to create streams.   Note that the Pravega controller startup routine involves more than exposing a TCP endpoint; some system streams are created before Pravega is truly ready.   Could the true readiness be detected somehow?

The solution will involve adding a `status` section to the CRD, and using the `UpdateStatus` method.

Note that the `status` section is expected to have some well-known elements, e.g. `conditions`.  The word _conditions_ has a meaning like "weather conditions" not "conditional logic".  There is a mindful design underlying this ([example](https://kubernetes.io/docs/concepts/architecture/#condition)).

Tristan1900:
### Change log description
- Add Conditions to Pravega cluster status.
- When cluster is ready, the `PodsReady` state will be set
### Purpose of the change
Fix #10 
### How to test the change
Tests should pass


pkg/apis/pravega/v1alpha1/status.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (ps *PravegaClusterStatus) ContainsCondition(condType PravegaClusterConditionType) bool {
+	if _, conditon := ps.getClusterCondition(condType); conditon != nil {
+		return true
+	}
+	return false
+}
+
+func (ps *PravegaClusterStatus) SetReadyCondition() {
+	c := newClusterCondition(PravegaClusterConditionReady, corev1.ConditionTrue, "Cluster available", "")
```
adrianmo: According to the [convention](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#typical-status-properties), `Reason` should be a one-word, CamelCase string. I would use `ClusterReady` to match the condition type.

pkg/apis/pravega/v1alpha1/status.go
```diff
@@ -0,0 +1,136 @@
...
+
+func newClusterCondition(condType PravegaClusterConditionType, status corev1.ConditionStatus, reason, message string) *PravegaClusterCondition {
+	now := time.Now().Format(time.RFC3339)
+	return &PravegaClusterCondition{
+		Type:               condType,
+		Status:             status,
+		Reason:             reason,
+		Message:            message,
+		LastUpdateTime:     now,
+		LastTransitionTime: now,
```
adrianmo: `LastTransitionTime` should only be set to `now` when the condition status changes from the previous value.

pkg/apis/pravega/v1alpha1/status.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+	return -1, nil
+}
+
+func (ps *PravegaClusterStatus) setClusterCondition(c PravegaClusterCondition) {
+	position, condition := ps.getClusterCondition(c.Type)
+	if condition != nil &&
+		condition.Status == c.Status && condition.Reason == c.Reason && condition.Message == c.Message {
+		return
```
adrianmo: If we don't set the new condition object, then the `LastUpdateTime` field will not be updated.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -297,6 +329,13 @@ func (r *ReconcilePravegaCluster) syncBookieSize(p *pravegav1alpha1.PravegaClust
 		if err != nil {
 			return fmt.Errorf("failed to sync pvcs of stateful-set (%s): %v", sts.Name, err)
 		}
+
+		p.Status.SetScalingCondition()
+		p.Status.ClearReadyCondition()
+		err = r.client.Update(context.TODO(), p)
```
adrianmo: We are updating the CR object from multiple places just to update the status information. We should try to unify the CR status update in one place, e.g., in the new `reconcileClusterReadiness` method.

Tristan1900: You are right.

pkg/apis/pravega/v1alpha1/status.go
Tristan1900: Got it

pkg/apis/pravega/v1alpha1/status.go
Tristan1900: Got it

pkg/apis/pravega/v1alpha1/status.go
Tristan1900: Got it. Thanks

Tristan1900:
LGTM! Thanks for the hard work!

adrianmo:
[Operator: Provide CR status information](https://trello.com/c/cVwUz0nj/111-operator-provide-cr-status-information)

PR #124, opened 2019-01-30
[WiP] Issue 81: Support rolling upgrades

adrianmo:
### Motivation

Pravega has been in active development for some time and versions have been released at a steady pace. However, there is no automated process, guidelines, or best practices to upgrade a Pravega deployment to a newer version.

### Limitations

Pravega is made up multiple components which can be differentiated into two blocks: internal and external components. Internal components refer to the Pravega Controller, the Segment Store, and BookKeeper. Whereas external components refer to ZooKeeper and the Tier 2 backend.

The Pravega operator scope is limited to internal components, and therefore, this upgrade procedure will only affect the aforementioned components. 

### Overview

The activity diagram below shows the overall upgrade process started by an end-user and performed by the operator.

pravega k8 upgrade 1

The user story would be the following:

- A user wants to upgrade a Pravega cluster named “pravega” from version X to version Y.
- The user expresses that desire by one of the following forms:
  - kubectl edit PravegaCluster pravega, and editing the “version” field in the YAML resource
  - kubectl patch PravegaCluster pravega -p '{"spec":{"version": ”Y”}}'
  - kubectl apply -f file.yaml, where file.yaml is the YAML resource definition with the “version” updated to “Y”
- The Pravega operator detects the version change and
  - if the upgrade is allowed, it applies the rolling upgrade
  - otherwise, it rejects the user request
- If the rolling upgrade faces any issues, the operator rolls back to the upgrade.

### Rolling upgrade process

pravega operator component update

Once an upgrade request has been validated, the operator will apply the rolling upgrade to all the components that are part of the Pravega cluster.

The order in which the components will be upgraded is the following:

1. Controller
2. Segment Store
3. BookKeeper

The upgrade workflow is as follows:

- The operator will change the PravegaCluster phase from RUNNING to UPGRADING to indicate that the cluster resource has an in-progress upgrade.
- For each cluster component, the operator will check if there is an ongoing rolling upgrade. To do that, the operator will check the rollout status for the Deployment (Pravega Controller) or StatefulSet (Pravega SegmentStore and BookKeeper).
  - If there’s an ongoing upgrade, the process will end and will start again at the next operator tick.
  - Otherwise, there’s no ongoing upgrade.
- The operator will check if the current component version matches the target version.
  - If no, the component is not upgraded or upgrading, so the operator will trigger a rolling upgrade using the “RollingUpdate” Kubernetes strategy by updating the container template in the Deployment or StatefulSet to use the new version. Then it will exit.
  - Otherwise, the component is already upgraded.
- The operator will check if there are more components to be upgraded.
  - If yes, it will go to 2.
  - Otherwise, all components are already upgraded
- The operator will change the PravegaCluster phase from UPGRADING to RUNNING to indicate that the cluster is already running with the new version.

#### Controller
TBD

#### Segment Store
TBD

#### Client
- Add ability to identify server-side component versions.
- Show warning logs if the client version does not match the server version, but it is compatible.
- Fail to start or throw exception if the client version is not compatible with the server version.

#### Operator
- Add support for `PravegaCluster` resource status to know whether or not the resource is ready to serve requests
  - https://github.com/pravega/pravega-operator/issues/10
- Implement an admission webhook to validate upgrade requests, i.e., whether the upgrade from version X to version Y is allowed/implemented or not.
  - https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks
  - https://github.com/kubernetes-sigs/controller-runtime/blob/master/example/validatingwebhook.go
  - https://github.com/pravega/pravega-operator/issues/7

### Version support

A Pravega cluster is upgraded as a whole. A user can’t upgrade only a singular component. 
Support rolling upgrade from one minor release to its immediate next minor release, e.g. from 0.5 to 0.6, no 0.5 to 0.7.
Support rolling upgrade from within the same minor release, e.g. from 0.5.0 to 0.5.4
Support rolling upgrade from any patch release to the immediate next minor release, e.g. from 0.5.6 to 0.6.0, or 0.5.2 to 0.6.4.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The Pravega operator manages Pravega clusters deployed to Kubernetes and automat
 
 - [x] Create and destroy a Pravega cluster
 - [x] Resize cluster
-- [ ] Rolling upgrades
+- [x] Rolling upgrades
 
 > Note that unchecked features are in the roadmap but not available yet.
 


adrianmo:
[Operator: Support Pravega upgrades](https://trello.com/c/htKcnzXO/10-operator-support-pravega-upgrades)

adrianmo:
@Tristan1900 can you please have a look at the PR? Thanks!

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -0,0 +1,372 @@
...
+	if sts.Status.UpdatedReplicas == sts.Status.Replicas &&
+		sts.Status.UpdatedReplicas == sts.Status.ReadyReplicas {
+		// StatefulSet upgrade completed
+		// TODO: wait until there is no under replicated ledger
+		// https://bookkeeper.apache.org/docs/4.7.2/reference/cli/#listunderreplicated
+		return true, nil
+	}
+
+	// Upgrade still in progress
+	// If all replicas are ready, upgrade an old pod
```
Tristan1900: I am wondering why Bookkeeper upgrading is using `onDelete` method instead of `rollingUpgrade` in SegmentStore?

pkg/controller/pravega/bookie.go
```diff
@@ -77,22 +80,27 @@ func MakeBookieStatefulSet(pravegaCluster *v1alpha1.PravegaCluster) *appsv1.Stat
...
+			Annotations: map[string]string{"pravega.version": p.Spec.Version},
 		},
-		Spec: *makeBookiePodSpec(pravegaCluster.Name, pravegaCluster.Spec.Bookkeeper),
+		Spec: *makeBookiePodSpec(p),
 	}
 }
 
-func makeBookiePodSpec(clusterName string, bookkeeperSpec *v1alpha1.BookkeeperSpec) *corev1.PodSpec {
+func makeBookiePodSpec(p *v1alpha1.PravegaCluster) *corev1.PodSpec {
+	terminationGracePeriodSeconds := int64(0)
```
Tristan1900: What is the reason for setting grace period seconds to be 0? Thanks

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -0,0 +1,372 @@
...
+
+	targetImage := fmt.Sprintf("%s:%s", p.Spec.Pravega.ImageRepository, p.Status.TargetVersion)
+
+	if deploy.Spec.Template.Spec.Containers[0].Image != targetImage {
+		// Need to update pod template
+		// This will trigger the rolling upgrade process
+		log.Printf("updating deployment (%s) pod template image to '%s'", deploy.Name, targetImage)
+
+		deploy.Spec.Template = pravega.MakeControllerPodTemplate(p)
+		// deploy.Spec.Template.Spec.Containers[0].Image = targetImage
```
Tristan1900: Not sure if we need to remove this comment?

pkg/controller/pravegacluster/upgrade.go
adrianmo: Good question. I'll create a document explaining the upgrade process and the reason behind those decisions.

pkg/controller/pravegacluster/upgrade.go
adrianmo: Yes, it has to be remoed. Good catch!

pkg/controller/pravega/bookie.go
adrianmo: Same as above, I'll create a document explaining the upgrade process and the reason behind those decisions :)

Tristan1900:
@adrianmo Thanks for updating the doc! LGTM.

adrianmo:
@Tristan1900 I'm re-requesting a review from you after the recent changes made to the PR. Thank you!

pkg/apis/pravega/v1alpha1/pravega.go
```diff
@@ -184,6 +191,27 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
 	return changed
 }
 
+// PravegaImageSpec defines the fields needed for a Pravega Docker image
+type PravegaImageSpec struct {
+	ImageSpec
+}
+
+func (s *PravegaImageSpec) withDefaults(c *ClusterSpec) (changed bool) {
```
Tristan1900: I think `c *ClusterSpec` is actually not used in this function.

pkg/apis/pravega/v1alpha1/bookkeeper.go
```diff
@@ -126,6 +133,27 @@ func (s *BookkeeperSpec) withDefaults() (changed bool) {
 	return changed
 }
 
+// BookkeeperImageSpec defines the fields needed for a BookKeeper Docker image
+type BookkeeperImageSpec struct {
+	ImageSpec
+}
+
+func (s *BookkeeperImageSpec) withDefaults(c *ClusterSpec) (changed bool) {
```
Tristan1900: Same here, argument seems not in use.

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
```diff
@@ -106,22 +106,26 @@ func (s *ClusterSpec) withDefaults() (changed bool) {
 
 	if s.Version == "" {
 		changed = true
-		s.Version = DefaultPravegaVersion
+		if s.Pravega != nil && s.Pravega.Image != nil && s.Pravega.Image.Tag != "" {
```
Tristan1900: Hmm, so if users don't specify `version`, instead they specify `pravega tag` and `bookkeeper tag`, the `pravega tag` will override the `bookkeeper tag`, which might make users confused. Can we deprecate `tag` in the future? We are still in alpha version, so in theory, we can break any API we want. What do you think?

pkg/apis/pravega/v1alpha1/pravega.go
adrianmo: Right. I've set the tag to `.spec.Version`.

pkg/apis/pravega/v1alpha1/bookkeeper.go
adrianmo: Right. I've set the tag to `.spec.Version`.

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
adrianmo: Right. One of the ideas behind this, apart from maintaining backwards compatibility, is that Pravega and BookKeeper images are both based on the same version to prevent users users from mistakenly using different versions of Pravega and BookKeeper, resulting in a potentially incompatible cluster. 

I added a deprecation note to the `Tag` definition in `pravegacluster_types.go`, and I'm also going to add it to the example CR.

Issue #126, opened 2019-02-13
Add an example script for cluster deployment


Tristan1900:
Add an example script and related files for cluster deployment to help users better understand the deployment procedure.

Tristan1900:
Will enhance the documentation instead.

PR #127, opened 2019-02-14
Issue 126: Add example scripts and files

Tristan1900:
### Change log description
- Add example scripts and files for Pravega cluster deployment.

### Purpose of the change
Solve #126 

### How to test the change
Run script should create a Pravega cluster automatically.

### DIFF
--- a/example/gke/install.sh
+++ b/example/gke/install.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Dell Inc., or its subsidiaries. All Rights Reserved.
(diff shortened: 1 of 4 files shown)


Tristan1900:
It seems that there is a chance that tests will fail. Will investigate further.

adrianmo:
@Tristan1900 I'm not sure we need this. Instead, we could enhance our documentation to install and use the operator in multiple Cloud platforms.

Tristan1900:
@adrianmo Sounds good! Close this now.

PR #129, opened 2019-03-04
Issue 130: Ability to specify container resources

### LINKED ISSUE #130: Set suitable CPU and memory resources to containers
Currently, the operator is not specifying the amount of compute resource that containers are allocated.

In Kubernetes, two type of resources can be specified:

- Resource requests: used by the scheduler to check if the Pod resource request can fit in a node depending on the available resources. If there's no node with enough resources to satisfy the request, the Pod will remain in a `Pending` state until it can be scheduled. By default, there's no request restriction.
- Resource limits: this limit is passed down to the container runtime, most likely Docker, to ensure that the container does not consume more resources than specified. If it does, it might be restarted or evicted.

More information on managing container resources: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container

In the operator scope, we need to:
- run some experiments and come up with adequate values for resource requests and limits to be used as default values.
- expose those settings to users on the API so that they can be configured to suit specific execution environments and resource restrictions/needs.

adrianmo:
### Change log description

This PR will bring the following changes:
- Set adequate default resource requests and limits for Pravega components
- Expose those resource setting to allow custom configurations
- Replace static `Xmx` JVM option with the dynamic, Docker-aware flags `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` which will set the heap memory size to the given Pod memory limit. (Ref: https://blogs.oracle.com/java-platform-group/java-se-support-for-docker-cpu-and-memory-limits)

At this moment, the PR is being used for testing with the goal of finding the suitable values for resource requests and limits.

### Purpose of the change
Fix #130 

### How to test the change
When defining the Pravega cluster specification in the YAML file, include the following options to set custom resource requests and limits for each Pravega component.

```
apiVersion: "pravega.pravega.io/v1alpha1"
kind: "PravegaCluster"
metadata:
  name: "example"
spec:
  ...
  bookkeeper:
    ...
    resources:
      requests:
        memory: "3Gi"
        cpu: "1000m"
      limits:
        memory: "5Gi"
        cpu: "2000m"

  pravega:
    ...
    controllerResources:
      requests:
        memory: "1Gi"
        cpu: "1000m"
      limits:
        memory: "3Gi"
        cpu: "2000m"

    segmentStoreResources:
      requests:
        memory: "3Gi"
        cpu: "1000m"
      limits:
        memory: "5Gi"
        cpu: "2000m"
```

You can verify that those values have been correctly set by looking at the Pods description.

---

### DIFF
--- a/example/cr-detailed.yaml
+++ b/example/cr-detailed.yaml
@@ -12,6 +12,13 @@ spec:
       pullPolicy: IfNotPresent
 
     replicas: 3
+    resources:
+      requests:
+        memory: "4Gi"
+        cpu: "4000m"
+      limits:
+        memory: "6Gi"
+        cpu: "6000m"
 
     storage:
       ledgerVolumeClaimTemplate:
@@ -34,7 +41,22 @@ spec:
 
   pravega:
     controllerReplicas: 1
+    controllerResources:
+      requests:
+        memory: "1Gi"
+        cpu: "1000m"
+      limits:
+        memory: "2Gi"
+        cpu: "2000m"
+
     segmentStoreReplicas: 3
+    segmentStoreResources:
+      requests:
+        memory: "4Gi"
+        cpu: "4000m"
+      limits:
+        memory: "6Gi"
+        cpu: "6000m"
 
     # Turn on Pravega Debug Logging
     debugLogging: false
--- a/example/cr-detailed.yaml
+++ b/example/cr-detailed.yaml
@@ -14,11 +14,11 @@ spec:
     replicas: 3
     resources:
       requests:
-        memory: "4Gi"
-        cpu: "4000m"
+        memory: "3Gi"
+        cpu: "1000m"
       limits:
-        memory: "6Gi"
-        cpu: "6000m"
+        memory: "5Gi"
+        cpu: "2000m"
 
     storage:
       ledgerVolumeClaimTemplate:
@@ -63,17 +63,17 @@ spec:
         memory: "1Gi"
         cpu: "1000m"
       limits:
-        memory: "2Gi"
+        memory: "3Gi"
         cpu: "2000m"
 
     segmentStoreReplicas: 3
     segmentStoreResources:
       requests:
-        memory: "4Gi"
-        cpu: "4000m"
+        memory: "3Gi"
+        cpu: "1000m"
       limits:
-        memory: "6Gi"
-        cpu: "6000m"
+        memory: "5Gi"
+        cpu: "2000m"
 
     # Turn on Pravega Debug Logging
     debugLogging: false
--- a/pkg/apis/pravega/v1alpha1/bookkeeper.go
+++ b/pkg/apis/pravega/v1alpha1/bookkeeper.go
@@ -110,12 +110,12 @@ func (s *BookkeeperSpec) withDefaults() (changed bool) {
 		changed = true
 		s.Resources = &v1.ResourceRequirements{
 			Requests: v1.ResourceList{
-				v1.ResourceCPU:    resource.MustParse("4000m"),
-				v1.ResourceMemory: resource.MustParse("4Gi"),
+				v1.ResourceCPU:    resource.MustParse("1000m"),
+				v1.ResourceMemory: resource.MustParse("3Gi"),
 			},
 			Limits: v1.ResourceList{
-				v1.ResourceCPU:    resource.MustParse("6000m"),
-				v1.ResourceMemory: resource.MustParse("6Gi"),
+				v1.ResourceCPU:    resource.MustParse("2000m"),
+				v1.ResourceMemory: resource.MustParse("5Gi"),
 			},
 		}
 	}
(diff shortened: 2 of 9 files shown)


pkg/apis/pravega/v1alpha1/bookkeeper.go
```diff
@@ -102,9 +106,24 @@ func (s *BookkeeperSpec) withDefaults() (changed bool) {
 		s.AutoRecovery = &boolTrue
 	}
 
+	if s.Resources == nil {
+		changed = true
+		s.Resources = &v1.ResourceRequirements{
+			Requests: v1.ResourceList{
+				v1.ResourceCPU:    resource.MustParse("500m"),
```
RaulGracia: Why these values are lower than the ones in `cr-detailed.yaml`? Shall we stick to a consistent defaults?

pkg/apis/pravega/v1alpha1/pravega.go
```diff
@@ -129,10 +137,39 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
...
+
 	if s.Tier2.withDefaults() {
 		changed = true
 	}
 
+	if s.ControllerResources == nil {
+		changed = true
+		s.ControllerResources = &v1.ResourceRequirements{
+			Requests: v1.ResourceList{
+				v1.ResourceCPU:    resource.MustParse("250m"),
```
RaulGracia: Same comment here with the defaults compared to `cr-detailed.yaml`.

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -151,44 +141,48 @@ func makeSegmentstorePodSpec(pravegaCluster *api.PravegaCluster) corev1.PodSpec
 	return podSpec
 }
 
-func MakeSegmentstoreConfigMap(pravegaCluster *api.PravegaCluster) *corev1.ConfigMap {
+func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
 	javaOpts := []string{
-		"-Xms1g -Xmx4g -XX:MaxDirectMemorySize=1g -Dpravegaservice.clusterName=" + pravegaCluster.Name,
+		"-Xms1g",
+		"-XX:+UnlockExperimentalVMOptions",
```
RaulGracia: Are these options configurable as part of another PR, right?

example/cr-detailed.yaml
```diff
@@ -12,6 +12,13 @@ spec:
       pullPolicy: IfNotPresent
 
     replicas: 3
+    resources:
```
RaulGracia: Is there any test that validates that these values are properly passed to the process?

pkg/apis/pravega/v1alpha1/bookkeeper.go
adrianmo: The default values are meant to represent the bare minimum requirements for Pravega to run. These values are intended to be use in a first contact with Pravega, functional testing, or development work. Therefore, we want to keep them low so that Pravega can be deployed to small Kubernetes clusters. Whereas the values in `cr-detailed.yaml` are just an example of how the default resources would be overridden to tune Pravega resources to a higher performance cluster.

pkg/apis/pravega/v1alpha1/pravega.go
adrianmo: Responded in https://github.com/pravega/pravega-operator/pull/129#discussion_r267449637

pkg/controller/pravega/pravega_segmentstore.go
adrianmo: No, JVM options are not externally configurable.

example/cr-detailed.yaml
adrianmo: No, we will work on adding a test to validate that resources are passed down to pod templates correctly.

Tristan1900: Test added. Could you please review it again? Thanks!

pkg/controller/pravega/pravega_segmentstore.go
RaulGracia: Ok, this may not be great, because we will rely on the default `-Xmx` value set by the JVM implementation. The default may be set as follows:
> Often its default value is 1/4th of your physical memory or 1GB (whichever is smaller).

So, we may end up providing 4GBs of memory to a pod, but 1GB at most will be devoted to the JVM (despite the main purpose of that pod is to run the JVM for the Pravega process at hand). What is you opinion about this? Have you used `-XX:+PrintFlagsFinal` to see what is the actual `-Xmx` value that is being used? Would it make sense to set the `-Xmx` value in function to the provisioned memory resources for the pod?

adrianmo:
> Would it make sense to set the `-Xmx` value in function to the provisioned memory resources for the pod?

@RaulGracia that's exactly what `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` flags do. These flags were introduced in Java 8u131 and Java 9 and, when they are set and `-Xmx` is not set, JVM heap size will be set according to the cgroup configuration, which is what we specify in the Pod resource limit.

PR #136, opened 2019-03-12
Update README: Add service account role definitions

adrianmo:
Add role and cluster role definitions when configuring Pravega cluster with external access enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,34 @@ spec:
 ...
 ```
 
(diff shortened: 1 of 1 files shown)


Tristan1900:
LGTM!
Just for a discussion. What do you think if we create such roles for users and create a rolebinding to grant access to their service account in the Operator when external access enabled?

adrianmo:
@Tristan1900 IMO, if a user wants to use a custom service account for Pravega components, they should make sure that those service accounts have all the required permissions stated in the documentation. In any case, let's further discuss this offline.

Issue #141, opened 2019-03-25
Bookie restarted with exception "java.lang.OutOfMemoryError: Java heap space"


vedanthh:
**While running moderate IO workload  in Longevity run (Total - 7 readers, 14 writers, 2500 events/sec, ~ 20 MB/s IO) observed multiple bookie (bookie1,bookie 2 and bookie3) restarted with exception java.lang.OutOfMemoryError: Java heap space on PKS environment**.

**Environment details: PKS / K8 with medium cluster:**
```
3 master: xlarge: 4 CPU, 16 GB Ram, 32 GB Disk
5 worker: 2xlarge: 8 CPU, 32 GB Ram, 64 GB Disk
Tier-1 storage is from VSAN datastore
Tier-2 storage curved on NFS Client provisioner using Isilon as backend
```

```
Pravega version: zk-closed-client-issue-0.5.0-2161.60655bf
Zookeeper Operator : pravega/zookeeper-operator:0.2.1
Pravega Operator: adrianmo/pravega-operator:issue-134-5
```
[bookie_controller_log.zip](https://github.com/pravega/pravega-operator/files/3002038/bookie_controller_log.zip)

```
NAME                                                 READY     STATUS    RESTARTS   AGE
pravega-bookie-0                                     1/1       Running   0          3d
pravega-bookie-1                                     1/1       Running   1          3d
pravega-bookie-2                                     1/1       Running   1          3d
pravega-bookie-3                                     1/1       Running   2          3d
pravega-operator-775554574d-7flrw                    1/1       Running   0          3d
pravega-pravega-controller-77cbcc59c7-hm9p9          1/1       Running   2          3d
pravega-pravega-controller-77cbcc59c7-rms82          1/1       Running   2          3d
pravega-pravega-segmentstore-0                       1/1       Running   1          3d
pravega-pravega-segmentstore-1                       1/1       Running   0          3d
pravega-pravega-segmentstore-2                       1/1       Running   1          3d
pravega-pravega-segmentstore-3                       1/1       Running   0          3d
pravega-zk-0                                         1/1       Running   0          3d
pravega-zk-1                                         1/1       Running   0          3d
pravega-zk-2                                         1/1       Running   0          3d
shaka-zulu-nfs-client-provisioner-59d7f8f84c-xvktb   1/1       Running   0          3d
zookeeper-operator-685bfcbbc5-tjtjr                  1/1       Running   0          3d
```

```
[...]

PR #145, opened 2019-03-25
Fix Pravega and BookKeeper Java options

### LINKED ISSUE #144: BookKeeper Java options are not properly set
When investigating #141 I found out that the BK Java process was not using the Java options set by the operator (https://github.com/pravega/pravega-operator/blob/0.3.1/pkg/controller/pravega/bookie.go#L206).

At the logs, I found the following error when the Bookies were starting:

```
/opt/bookkeeper/conf/bkenv.sh: line 35: -XX:+UnlockExperimentalVMOptions: command not found
```

Wrapping those options in double quotes fixed the issue and Bookies were able to pick up the Java extra options set by the operator, however, I inspected the process running inside the Docker container and found out conflicting and duplicated options.

Excerpt of running `ps aux` inside a Bookie container: 
```
...
-Djava.net.preferIPv4Stack=true
-Xms1g
-Xmx1g
-XX:MaxDirectMemorySize=2g
-XX:+UseG1GC
-XX:MaxGCPauseMillis=10
-XX:+ParallelRefProcEnabled
-XX:+UnlockExperimentalVMOptions
-XX:+AggressiveOpts
-XX:+DoEscapeAnalysis
-XX:ParallelGCThreads=32
-XX:ConcGCThreads=32
-XX:G1NewSizePercent=50
-XX:+DisableExplicitGC
-XX:-ResizePLAB
-XX:+PrintGCDetails
-XX:+PrintGCApplicationStoppedTime
-XX:+UseGCLogFileRotation
-XX:NumberOfGCLogFiles=5
-XX:GCLogFileSize=64m
-Xloggc:/opt/bookkeeper/logs/gc_%p.log
-Dio.netty.leakDetectionLevel=disabled
-Dio.netty.recycler.maxCapacity.default=1000
-Dio.netty.recycler.linkCapacity=1024
-Dlog4j.configuration=log4j.properties
-Dbookkeeper.root.logger=INFO,CONSOLE
-Dbookkeeper.log.dir=/opt/bookkeeper/logs
-Dbookkeeper.log.file=bookkeeper-server.log
-Xms1g
-XX:+UnlockExperimentalVMOptions
-XX:+UseCGroupMemoryLimitForHeap
-XX:MaxRAMFraction=1
-XX:MaxDirectMemorySize=1g
-XX:+UseG1GC
-XX:MaxGCPauseMillis=10
-XX:+ParallelRefProcEnabled
-XX:+AggressiveOpts
-XX:+DoEscapeAnalysis
-XX:ParallelGCThreads=32
-XX:ConcGCThreads=32
-XX:G1NewSizePercent=50
-XX:+DisableExplicitGC
-XX:-ResizePLAB
...
```

For example, the operator is using `-XX:+UnlockExperimentalVMOptions` and `-XX:+UseCGroupMemoryLimitForHeap` to dynamically set the JVM heap size to the container memory limit, however, `-Xmx1g` is overriding that behavior and hardcoding the heap size to 1 GB.

After further investigation, I found that those conflicting options are set by Bookkeeper itself in https://github.com/apache/bookkeeper/blob/master/bin/common.sh#L99. Based on that script, we should be using the `BOOKIE_MEM_OPTS`, `BOOKIE_GC_OPTS`, and `BOOKIE_GC_LOGGING_OPTS` environment variables to override the default values. instead of setting them using the `BK_BOOKIE_EXTRA_OPTS` variable.

adrianmo:
### Change log description

- Set BookKeeper Java options using the `BOOKIE_MEM_OPTS`, `BOOKIE_GC_OPTS`, and `BOOKIE_GC_LOGGING_OPTS` environment variables instead of `BK_BOOKIE_EXTRA_OPTS`.
- Make Segmentstore, Controller, and BookKeeper processes exit and crash on out of memory errors.
- Fix CI triggers.

### Purpose of the change
Fix #144
Related to #141 

### How to test the change

After deploying a Pravega cluster, `kubectl exec` into the containers, list the processes with `ps aux`, and check that Java options are properly passed to the Java process. 

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
 
     - stage: e2e
       name: End-to-end tests
-      if: type = pull_request OR branch = master
+      if: type = pull_request OR branch = master OR tag IS present
       services:
         - docker
       env:
@@ -85,7 +85,7 @@ jobs:
 
     - stage: deploy
       name: Push Docker image
-      if: branch = master AND tag IS present
+      if: type != pull_request AND tag IS present
       services:
         - docker
       script:
--- a/pkg/controller/pravega/bookie.go
+++ b/pkg/controller/pravega/bookie.go
@@ -184,15 +184,18 @@ func makeBookieVolumeClaimTemplates(spec *v1alpha1.BookkeeperSpec) []corev1.Pers
 }
 
 func MakeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.ConfigMap {
-	javaOpts := []string{
-		// Leading and trailing double quotes are required for Bookies to
-		// pick up these Java options.
-		"\"",
+	memoryOpts := []string{
 		"-Xms1g",
 		"-XX:+UnlockExperimentalVMOptions",
 		"-XX:+UseCGroupMemoryLimitForHeap",
 		"-XX:MaxRAMFraction=1",
 		"-XX:MaxDirectMemorySize=1g",
+		"-XX:+ExitOnOutOfMemoryError",
+		"-XX:+CrashOnOutOfMemoryError",
+		"-XX:+HeapDumpOnOutOfMemoryError",
+	}
+
+	gcOpts := []string{
 		"-XX:+UseG1GC",
 		"-XX:MaxGCPauseMillis=10",
 		"-XX:+ParallelRefProcEnabled",
@@ -203,12 +206,21 @@ func MakeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.Config
 		"-XX:G1NewSizePercent=50",
 		"-XX:+DisableExplicitGC",
 		"-XX:-ResizePLAB",
-		"\"",
+	}
+
+	gcLoggingOpts := []string{
+		"-XX:+PrintGCDetails",
+		"-XX:+PrintGCApplicationStoppedTime",
+		"-XX:+UseGCLogFileRotation",
+		"-XX:NumberOfGCLogFiles=5",
+		"-XX:GCLogFileSize=64m",
 	}
 
 	configData := map[string]string{
-		"BK_BOOKIE_EXTRA_OPTS": strings.Join(javaOpts, " "),
-		"ZK_URL":               pravegaCluster.Spec.ZookeeperUri,
+		"BOOKIE_MEM_OPTS":        strings.Join(memoryOpts, " "),
+		"BOOKIE_GC_OPTS":         strings.Join(gcOpts, " "),
+		"BOOKIE_GC_LOGGING_OPTS": strings.Join(gcLoggingOpts, " "),
+		"ZK_URL":                 pravegaCluster.Spec.ZookeeperUri,
 		// Set useHostNameAsBookieID to false until BookKeeper Docker
 		// image is updated to 4.7
 		// This value can be explicitly overridden when using the operator
--- a/pkg/controller/pravega/bookie.go
+++ b/pkg/controller/pravega/bookie.go
@@ -185,6 +185,9 @@ func makeBookieVolumeClaimTemplates(spec *v1alpha1.BookkeeperSpec) []corev1.Pers
 
 func MakeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.ConfigMap {
 	javaOpts := []string{
+		// Leading and trailing double quotes are required for Bookies to
+		// pick up these Java options.
+		"\"",
 		"-Xms1g",
 		"-XX:+UnlockExperimentalVMOptions",
 		"-XX:+UseCGroupMemoryLimitForHeap",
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -85,7 +85,7 @@ jobs:
 
     - stage: deploy
       name: Push Docker image
-      if: branch = master AND tag IS present
+      if: type != pull_request AND tag IS present
```
Tristan1900: I'm seeing similar config [here](https://github.com/operator-framework/operator-sdk/blob/master/.travis.yml#L50) but I don't understand the reason. Does this mean we can push images to the docker hub even in other branches and why do we want to do that? Thanks!

pkg/controller/pravega/bookie.go
```diff
@@ -202,9 +208,19 @@ func MakeBookieConfigMap(pravegaCluster *v1alpha1.PravegaCluster) *corev1.Config
 		"-XX:-ResizePLAB",
 	}
 
+	gcLoggingOpts := []string{
+		"-XX:+PrintGCDetails",
```
Tristan1900: Probably better to add a timestamp  `-XX:+PrintGCDateStamps`  and a log dir, e.g. `-Xloggc:/tmp/gclog.log`, see [this](https://dzone.com/articles/enabling-and-analysing-the-garbage-collection-log). what do you think?

pkg/controller/pravega/pravega_controller.go
```diff
@@ -120,6 +120,10 @@ func MakeControllerConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
 		"-Xms1g",
 		"-XX:+UnlockExperimentalVMOptions",
 		"-XX:+UseCGroupMemoryLimitForHeap",
+		"-XX:+ExitOnOutOfMemoryError",
+		"-XX:+CrashOnOutOfMemoryError",
+		"-XX:+HeapDumpOnOutOfMemoryError",
+		"-XX:MaxRAMFraction=1",
```
Tristan1900: One potential problem for `-XX:MaxRAMFraction=1`  is discussed [here](https://medium.com/adorsys/jvm-memory-settings-in-a-container-environment-64b0840e1d9e#fn10-20439).
The risk of setting this to be 1 is that when under heavy load containers can get killed by k8s because of JVM tries to use up all the RAM for heap. There are other processes in the container as well as JVM overhead, so it can cause container to be OOM.
One solution is to set it to be 2, the disadvantage is that the mem utilization is 50%.
The better solution is to use `XX:MaxRAMPercentage `, the problem is that it only supported in Java 10 and above.

.travis.yml
adrianmo: This means that we only want to push images when we are not in a PR and a tag is present, i.e., when we create a GitHub release. The previous condition was never met because when a tag is present, the branch is never `master`, instead, the branch is equal to the tag name. For example, [this build](https://travis-ci.org/pravega/pravega-operator/builds/509965043) should have triggered a deploy, but it didn't.

pkg/controller/pravega/pravega_controller.go
adrianmo: Good catch @Tristan1900. Actually, we have already seen _OOMKilled_ (ref: https://github.com/pravega/pravega-operator/issues/141) when using default resource limits. My take is that we set `XX:MaxRAMFraction` to `2` until Pravega supports Java 10+. It's not optimal, but at least it should prevent pods from being _OOMKilled_ when they are under heavy load.

pkg/controller/pravega/bookie.go
adrianmo: Yes, I'll include the `-XX:+PrintGCDateStamps` option. About logging to a file, I don't think we should do that in Kubernetes, because the root file system is ephemeral and we won't be able to access the log file if the container dies. Instead, we want GC details to be printed to stdout so that it's handled by the logging agent (e.g. fluentd), which is the default behavior.

.travis.yml
Tristan1900: Got it, Thanks.

pkg/controller/pravega/bookie.go
Tristan1900: Oh, so the default output is stdout. That makes more sense.

pkg/controller/pravega/pravega_controller.go
Tristan1900: Thanks! That sounds good to me.

Tristan1900: Just to make sure. To prevent `java.lang.OutOfMemoryError: Java heap space` in #141, we still need to set memory limit high enough. Setting XX:MaxRAMFraction to 2 only guarantee containers don't get killed by k8s silently.

adrianmo:
Waiting on @vedanthh confirmation in #141.

adrianmo:
Got @vedanthh confirmation in https://github.com/pravega/pravega-operator/issues/141#issuecomment-477457356

PR #159, opened 2019-04-05
Issue 7, 156: Admission webhook

### LINKED ISSUE #7: Admission Controller for PravegaCluster resource

### LINKED ISSUE #156: Validate upgrade requests
At this moment, upgrade requests are not being validated, meaning that the operator will always trigger an upgrade process to any version specified in the resource definition.

We need to check if the version is explicitly supported. The operator should not attempt to upgrade to a non-supported or non-compatible version. Upgrade support to a particular version needs to be implemented and added to the operator. Furthermore, some versions might require specific actions and verifications, like applying migrations. 

This could be implemented as an admission webhook as part of #7. 

Related documentation:
- https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks
- https://github.com/kubernetes-sigs/controller-runtime/blob/master/example/validatingwebhook.go
#7

Tristan1900:
### Change log description
- Add admission webhook that supports version validation
- Add unit test
- Add e2e test

### What the code does 

There are two concepts here, the admission webhook server and the admission webhook. Admission webhook server is like an http server that serves the incoming requests. Admission webhook implements the application logic and register itself to the webhook server, there are two kinds of admission webhook, `mutating admission webhook` and `validating admission Webhook`, right now we only use `mutating admission Webhook`. The admission  webhook server starts in the Operator pod with a fronting k8s `service`. When requests come into k8s, apiserver will route the request to that `service` and the backend admission webhook server will get that request and call the admission webhook for processing. The port that admission webhook server listens to is port 443. To bind port 443, we need to grant permission to Operator process using `setcap CAP_NET_BIND_SERVICE=+eip`.

There is a supported version list in our webhook. Any requests with an invalid Pravega version will be rejected. For example `0.6.0` is an invalid Pravega version because there is no such released version yet. There is also a supported upgrade path in our webhook, it will reject the request if the upgrade is not supported. For example, rolling upgrade from `0.3.0` to `0.4.0` is not supported. However, any upgrade to patch version is supported, such as `0.4.0` to `0.4.1`.

### Purpose of the change
- Fix #7 
- Fix #156 

### How to test the change
Tests should pass

---


Dockerfile
```diff
@@ -28,10 +28,15 @@ RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o ${GOBIN}/${PROJECT_NAME} \
...
 
+RUN apk add --update \
+    sudo \
+    libcap
+
 ARG PROJECT_NAME=pravega-operator
 ARG REPO_PATH=github.com/pravega/$PROJECT_NAME
 
 COPY --from=go-builder ${GOBIN}/${PROJECT_NAME} /usr/local/bin/${PROJECT_NAME}
+RUN sudo setcap CAP_NET_BIND_SERVICE=+eip /usr/local/bin/${PROJECT_NAME}
```
adrianmo: @Tristan1900 can you explain why we need the additional dependencies and `setcap` command?

deploy/operator.yaml
```diff
@@ -33,3 +34,13 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "pravega-operator"
+            - name: SECRET_NAME
```
adrianmo: why `SECRET_NAME`? I see that is being used as the webhook server name, but not sure why we can't use the operator name.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,213 @@
...
+	pravega := &pravegav1alpha1.PravegaCluster{}
+
+	if err := pwh.decoder.Decode(req, pravega); err != nil {
+		return admission.ErrorResponse(http.StatusBadRequest, err)
+	}
+	copy := pravega.DeepCopy()
+
+	err := pwh.validatePravegaManifest(ctx, copy)
+	if err != nil {
+		return admission.ErrorResponse(http.StatusInternalServerError, err)
```
adrianmo: Not sure we should return a 500 Error if validation fails. E.g. if the user wants to use a non-supported Pravega version, the server should return a 400 BadRequest response. However, if there's an unexpected error during the validation, 500 Error is fine. 

Maybe we could make the `validatePravegaManifest` func to return `(bool, error)`, the `bool` value would indicate whether or not the request is valid, if `false` the server will return 400 BadRequest.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return admission.ValidationResponse(true, "")
+}
+
+func (pwh *pravegaWebhookHandler) validatePravegaManifest(ctx context.Context, p *pravegav1alpha1.PravegaCluster) error {
+	if error := pwh.validatePravegaVersion(ctx, p); error != nil {
+		return fmt.Errorf("Failed to pass version validation: %v", error)
+	}
+
+	//TODO: Other validations
```
adrianmo: Change comment to `// Add other validators here`

Dockerfile
Tristan1900: Sure! Webhook server needs to listen to 443 port for any requests from Apiserver. By default, it seems the operator has no permission to use the restricted port. `setcap` can grant permission that the operator needs.

pkg/util/pravegacluster.go
```diff
@@ -198,6 +198,18 @@ func CompareVersions(v1, v2, operator string) (bool, error) {
 	return constraints.Check(clusterVersion), nil
 }
 
+func ContainsVersion(list []string, version string) bool {
+	result := false
+	for _, v := range list {
+		if match, _ := CompareVersions(version, v, "="); match {
```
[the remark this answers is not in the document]
Tristan1900: You are right, Thanks!

deploy/operator.yaml
Tristan1900: Hmm. I think the server name is `admission-webhook-server`
So the reason for passing this as an environment variable is to let webhook knows the `secret` name that it can use. But I just find that we may skip the `secret` part, I will update the change later.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,213 @@
...
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	CertDir = "/var/run/secrets/kubernetes.io/"
+)
+
+var (
+	compatibilityMatrix = []string{
```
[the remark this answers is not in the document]
Tristan1900: Thanks, you are right!

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,213 @@
...
+func newWebhookServer(mgr manager.Manager) (*webhook.Server, error) {
+	return webhook.NewServer("admission-webhook-server", mgr, webhook.ServerOptions{
+		CertDir: CertDir,
+		BootstrapOptions: &webhook.BootstrapOptions{
+			Secret: &types.NamespacedName{
+				Namespace: os.Getenv("WATCH_NAMESPACE"),
+				Name:      os.Getenv("SECRET_NAME"),
+			},
+
+			// TODO: garbage collect webhook service
```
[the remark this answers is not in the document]
Tristan1900: Oh sorry for my bad comment. To create the webhook server we need to create a k8s `service` in front of it. Here I mean that k8s `service` is not garbage collected. The webhook server dies with our operator.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,213 @@
...
+			Service: &webhook.Service{
+				Namespace: os.Getenv("WATCH_NAMESPACE"),
+				Name:      "admission-webhook-server-service",
+				Selectors: map[string]string{
+					"component": "operator",
```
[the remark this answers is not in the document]
Tristan1900: This is for k8s `service` to select its backend, which is our operator pod. 
 How about `component: pravega-operaor`?

pkg/webhook/webhook.go
Tristan1900: Right! Make sense. Thanks for the comment.

pkg/webhook/webhook.go
Tristan1900: Got it.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,201 @@
...
+
+	//TODO: Add other validators here
+	return "200", nil
+}
+
+func (pwh *pravegaWebhookHandler) validatePravegaVersion(ctx context.Context, p *pravegav1alpha1.PravegaCluster) (string, error) {
+	// Check if the request has a legal Pravega version
+	if p.Spec.Version == "" {
+		if p.Spec.Pravega != nil && p.Spec.Pravega.Image != nil && p.Spec.Pravega.Image.Tag != "" {
+			p.Spec.Version = p.Spec.Pravega.Image.Tag
```
adrianmo: According to the [documentation](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) `ValidatingAdmissionWebhook` should not modify the object. `MutatingAdmissionWebhook` are used for that. 

`MutatingAdmissionWebhook` run before `ValidatingAdmissionWebhook`, so maybe we can create a `MutatingAdmissionWebhook` that just applies the default values to the object, i.e. `p.withDefaults()`. 

What do you think?

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return admission.ValidationResponse(true, "")
+}
+
+func (pwh *pravegaWebhookHandler) validatePravegaManifest(ctx context.Context, p *pravegav1alpha1.PravegaCluster) (string, error) {
+	if code, error := pwh.validatePravegaVersion(ctx, p); code != "200" {
+		return code, error
+	}
+
+	//TODO: Add other validators here
```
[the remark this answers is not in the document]
Tristan1900: Oh got it!

pkg/webhook/webhook.go
Tristan1900: Sorry I will make it more clear!

Tristan1900:
Thanks @adrianmo for the comments? Could you please review it again? Thanks!

Tristan1900:
@adrianmo Thanks for pointing that out! It makes sense to disable webhook when using `up local`. I made the change last Friday to address this but forgot to reply to your comments. Sorry about that. I see you've merged this PR, thanks for the review!

PR #165, opened 2019-04-17
Issue 163: Create Helm chart to install Pravega

### LINKED ISSUE #163: Create Helm chart for PravegaCluster
We currently have a Helm chart for deploying the operator, but we are missing a full-featured chart that creates all needed resources and deploys a Pravega cluster. The resources that this chart needs to manage are the following:

- PravegaCluster CRD
- Operator deployment
- Operator service account
- Operator role
- Operator clusterrole
- Operator rolebinding
- Operator clusterrolebinding
- PravegaCluster CR

All these resources can be currently found at the `deploy/` directory. These manifests need to be transformed into chart templates to make them configurable. 

As a first version of the chart we don't need to make it fully configurable, but at least we need to be able to configure the following options:

- cluster name
- version
- image repository
- replicas for BK, SS, and C

As a reference, we can take a look at other Helm charts in https://github.com/helm/charts/tree/master/stable.

adrianmo:
### Change log description

- [x] Update existing `pravega-operator` Helm chart
- [x] Add documentation for `pravega-operator` chart
- [x] Add tests for `pravega-operator` chart 
- [x] Create a `pravega` Helm chart to install a Pravega cluster with initial set of configuration parameters
- [x] Add documentation for `pravega` chart
- [x] Add tests for `pravega` charts

### Purpose of the change
- Fix #163 

### How to test the change
- Follow the instructions in the charts directory to install and check that the charts works as expected
- Tests should pass

---


charts/pravega-operator/templates/operator.yaml
```diff
@@ -1,21 +1,38 @@
+{{- if .Values.pravegaOperator.create }}
```
adrianmo: Does this chart make sense at all if it doesn't install the operator?

Tristan1900:
@adrianmo Great suggestion! Will do that.

charts/pravega-operator/templates/operator.yaml
Tristan1900: Right! Forgot to delete that, it used to be the Pravega chart.

Tristan1900:
LGTM!

PR #190, opened 2019-05-20
Issue 187: Webhook rejects requests when cluster is upgrading

### LINKED ISSUE #187: "admission webhook" should restrict to modify the PravegaCluster "version" field if there is an active upgrade exists.
**"admission webhook" should restrict to modify the PravegaCluster "version" field if there is an active upgrade process is going on.**

Right now, it is allowed to modify the "version" field of PravegaCluster through `kubectl edit PravegaCluster` when there is an active upgrade is going on but that changes never kicks off the new upgrade process since there is one already exists.

Pravega-operator version: `0.4.0-rc1`

Tristan1900:
### Change log description
Enable webhook to reject requests when cluster is upgrading

### Purpose of the change
Fix #187

### How to test
Unit test should pass

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -63,7 +63,7 @@ func (pwh *pravegaWebhookHandler) Handle(ctx context.Context, req admissiontypes
 	copy := pravega.DeepCopy()
 
 	if err := pwh.clusterIsAvailable(ctx, copy); err != nil {
-		return admission.ErrorResponse(http.StatusBadRequest, err)
+		return admission.ErrorResponse(http.StatusServiceUnavailable, err)
 	}
 
 	if err := pwh.mutatePravegaManifest(ctx, copy); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/webhook/webhook.go
```diff
@@ -60,6 +62,10 @@ func (pwh *pravegaWebhookHandler) Handle(ctx context.Context, req admissiontypes
 	}
 	copy := pravega.DeepCopy()
 
+	if err := pwh.clusterIsAvailable(ctx, copy); err != nil {
+		return admission.ErrorResponse(http.StatusBadRequest, err)
```
adrianmo: I believe that, in this case, we should return a server error (5xx) instead of a user error (4xx). `BadRequest` is telling the user that the request is malformed, which is not the case, it's just that the cluster status does not allow this type of request temporarily.

Tristan1900: Got it!

PR #229, opened 2019-07-11
Issue 228: Use emptyDir as default cache volume

### LINKED ISSUE #228: Segmentstore cache should use local disk if PVC is not specified
### Description
Right now the Segmentstore cache is using a persistent volume to store data. If the pvc template is not specified, the operator will set a default pvc template. However, according to this [line](https://github.com/pravega/pravega-operator/blob/master/pkg/apis/pravega/v1alpha1/pravega.go#L91), a `emptyDir` should be used if a template is not provided.

Also, local disk tends to have better latency performance than remote disk, one might want to just use `emptyDir` in some situations. 

### Importance
Potential blocker to various issues in Pravega

### Location
Segmentstore deployment code and api code.

If the pvc template is not provided in pravega manifest, use an `emptyDir` as the segmentstore cache volume.

Tristan1900:
### Change log description
This PR uses the `emptyDir` as the default volume for segmentstore cache.

### Purpose of the change
Fix #228 

### What the code does
Remove the default cache template in code and specify the `emptyDir` when cache template is nil. 

### How to verify it
When segmentstore pvc template is not specified, the cache volume should be the `emptyDir`. We can log into the segmentstore pod and do `df` to check the volume size. 

### DIFF
--- a/pkg/apis/pravega/v1alpha1/pravega.go
+++ b/pkg/apis/pravega/v1alpha1/pravega.go
@@ -139,18 +139,6 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
 		s.Options = map[string]string{}
 	}
 
-	if s.CacheVolumeClaimTemplate == nil {
-		changed = true
-		s.CacheVolumeClaimTemplate = &v1.PersistentVolumeClaimSpec{
-			AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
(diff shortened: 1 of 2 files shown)


RaulGracia:
@Tristan1900 I think that we can close this PR, do you agree?

Tristan1900:
@RaulGracia Agreed. This PR is not necessary anymore.

PR #240, opened 2019-07-25
Issue 211: Adding annotation to segment store service for external dns

### LINKED ISSUE #211: Support for External DNS names
### Description
The goal is to assign an external DNS name to each segment store, compatible with the [External-DNS component](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/faq.md#how-do-i-specify-a-dns-name-for-my-kubernetes-objects).   Given a domain suffix for the cluster, each segment store should be assigned a reasonable name, e.g. `<name>.<domain>` where `<name>` is the segment store pod name, and `<domain>` is a value provided by the `PravegaCluster` spec.  The assigned name should become the 'advertised' address of the segment store for use by the Pravega client.

The motivation is to support TLS connections to segment store with hostname verification enabled.  Related to pravega/pravega#3899.

Approach:
1. add a new field to the `PravegaCluster` spec, e.g. `ingress/domain`.
2. When creating a `Service` for a given `SegmentStore`, compute an FQDN as follows: `<name>.<domain>.` (note the period after the suffix).  
3. Record the FQDN as an `annotation` on the `Service` resource.  The annotation should have the key `external-dns.alpha.kubernetes.io/hostname`, to be compatible with the External-DNS component.
4. Adjust the Pravega entrypoint script to pick up the annotation when present (see pravega/pravega#3899).

The expected result is:
1. the External-DNS component creates a DNS record for the segment store.
2. the segment store advertises the correct DNS name to clients.
3. client is able to make a TLS connection with hostname verification enabled.

pbelgundi:

### Change log description
Added external-dns annotation to segment store external service that can be used by external-dns provider  to assign a DNS name to the service.
### Purpose of the change
Fixes #211

### What the code does
Added `domainName` to Pravega Cluster Spec, that can be set via manifest to arrive at the FQDN for a single segment store node.
FQDN for segment store pod = SegmentStore_Pod_Name.domainName. ( ends with a . )

Added annotation `external-dns.alpha.kubernetes.io/hostname` to external service for each segment store container.

If external access=true, but domain name is not set in manifest, it would default to "pravega.io"

### How to verify it
1. In Pravega spec set externalaccess=true, and domainName=`example.com.
2 `kubectl svc pravega-pravega-segmentstore-0 -o yaml` should show :
 `external-dns.alpha.kubernetes.io/hostname: pravega-pravega-segmentstore-0.example.com.` 
3. Check External-DNS component creates a DNS record for the segment store.

### DIFF
--- a/pkg/apis/pravega/v1alpha1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1alpha1/pravegacluster_types.go
@@ -25,9 +25,6 @@ const (
 	// DefaultPravegaVersion is the default tag used for for the Pravega
 	// Docker image
 	DefaultPravegaVersion = "0.4.0"
-
-	// Default Domain Name
-	DefaultDomainName = "pravega.io"
 )
 
 func init() {
@@ -170,10 +167,6 @@ func (e *ExternalAccess) withDefaults() (changed bool) {
 		if e.Type == "" {
 			changed = true
 			e.Type = DefaultServiceType
-		}
-		if e.DomainName == "" {
-			changed = true
-			e.DomainName = DefaultDomainName
 		}
 	}
 	return changed
--- a/pkg/apis/pravega/v1alpha1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1alpha1/pravegacluster_types.go
@@ -25,6 +25,9 @@ const (
 	// DefaultPravegaVersion is the default tag used for for the Pravega
 	// Docker image
 	DefaultPravegaVersion = "0.4.0"
+
+	// Default Domain Name
+	DefaultDomainName = "pravega.io"
 )
 
 func init() {
@@ -152,17 +155,27 @@ type ExternalAccess struct {
 	// Options are "LoadBalancer" and "NodePort".
 	// By default, if external access is enabled, it will use "LoadBalancer"
 	Type v1.ServiceType `json:"type,omitempty"`
+
+	// Domain Name to be used for External Access
+	// This value is ignored if External Access is disabled
+	DomainName string `json:"domainName,omitempty"`
 }
 
 func (e *ExternalAccess) withDefaults() (changed bool) {
-	if e.Enabled == false && e.Type != "" {
+	if e.Enabled == false && (e.Type != "" || e.DomainName != "") {
 		changed = true
 		e.Type = ""
-	} else if e.Enabled == true && e.Type == "" {
-		changed = true
-		e.Type = DefaultServiceType
+		e.DomainName = ""
+	} else if e.Enabled == true {
+		if e.Type == "" {
+			changed = true
+			e.Type = DefaultServiceType
+		}
+		if e.DomainName == "" {
+			changed = true
+			e.DomainName = DefaultDomainName
+		}
 	}
-
 	return changed
 }
 
--- a/pkg/controller/pravega/pravega_segmentstore.go
+++ b/pkg/controller/pravega/pravega_segmentstore.go
@@ -335,16 +335,22 @@ func MakeSegmentStoreExternalServices(pravegaCluster *api.PravegaCluster) []*cor
 	var service *corev1.Service
 	var ssPodName string
 	var ssFQDN string
+	var annotationMap map[string]string
 
-	domainName := strings.TrimSpace(pravegaCluster.Spec.ExternalAccess.DomainName)
 	services := make([]*corev1.Service, pravegaCluster.Spec.Pravega.SegmentStoreReplicas)
 
 	for i := int32(0); i < pravegaCluster.Spec.Pravega.SegmentStoreReplicas; i++ {
 		ssPodName = util.ServiceNameForSegmentStore(pravegaCluster.Name, i)
-		if strings.HasSuffix(domainName, dot) {
-			ssFQDN = ssPodName + dot + domainName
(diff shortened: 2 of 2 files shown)


pbelgundi:
Service shows annotation set correctly:
```
$kubectl get svc pravega-pravega-segmentstore-0 -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    external-dns.alpha.kubernetes.io/hostname: pravega-pravega-segmentstore-0.example.com.
    ncp/internal_ip_for_policy: 100.64.65.241
  creationTimestamp: "2019-07-25T12:10:39Z"
 . . .
```

pbelgundi:
When external_access = true, but domainName is not set in manifest, default value is used:
```
$ kubectl get svc pravega-pravega-segmentstore-0 -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    external-dns.alpha.kubernetes.io/hostname: pravega-pravega-segmentstore-0.pravega.io.
    ncp/internal_ip_for_policy: 100.64.65.241
  creationTimestamp: "2019-07-25T12:21:46Z"
```

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -329,18 +333,31 @@ func MakeSegmentStoreHeadlessService(pravegaCluster *api.PravegaCluster) *corev1
...
+
+	domainName := strings.TrimSpace(pravegaCluster.Spec.ExternalAccess.DomainName)
 	services := make([]*corev1.Service, pravegaCluster.Spec.Pravega.SegmentStoreReplicas)
 
 	for i := int32(0); i < pravegaCluster.Spec.Pravega.SegmentStoreReplicas; i++ {
+		ssPodName = util.ServiceNameForSegmentStore(pravegaCluster.Name, i)
+		if strings.HasSuffix(domainName, dot) {
+			ssFQDN = ssPodName + dot + domainName
+		} else {
+			ssFQDN = ssPodName + dot + pravegaCluster.Spec.ExternalAccess.DomainName + dot
```
EronWright: Shouldn't this read:
```
ssFQDN = ssPodName + dot + domainName + dot
```

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -329,18 +333,31 @@ func MakeSegmentStoreHeadlessService(pravegaCluster *api.PravegaCluster) *corev1
...
 				Kind:       "Service",
 				APIVersion: "v1",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      util.ServiceNameForSegmentStore(pravegaCluster.Name, i),
+				Name:      ssPodName,
 				Namespace: pravegaCluster.Namespace,
 				Labels:    util.LabelsForSegmentStore(pravegaCluster),
+				Annotations: map[string]string{
+					externalDNSAnnotationKey: ssFQDN,
```
EronWright: To eliminate the default suffix of `pravega.io`, compute an `annotations` map above, which may or may not contain an external DNS annotation.   Then, on this line, you can simply use the map.

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
```diff
@@ -25,6 +25,9 @@ const (
 	// DefaultPravegaVersion is the default tag used for for the Pravega
 	// Docker image
 	DefaultPravegaVersion = "0.4.0"
+
+	// Default Domain Name
+	DefaultDomainName = "pravega.io"
```
EronWright: Don't use this as a default value.   Instead, conditionally apply the external DNS annotation based on whether a domain was provided.

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: Sure. Will change this

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: I think we would need to have a default 'domainname' value to handle the case where external access is enabled but domainname is not set by the user in the manifest, especially since this value is also going to be used in "init_kubernetes.sh" for advertising SS service. The other option is to throw an error when that happens.

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
pbelgundi: Sure.

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: Ok.

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: Done

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -329,18 +333,35 @@ func MakeSegmentStoreHeadlessService(pravegaCluster *api.PravegaCluster) *corev1
 
 func MakeSegmentStoreExternalServices(pravegaCluster *api.PravegaCluster) []*corev1.Service {
 	var service *corev1.Service
+	var ssPodName string
+	var ssFQDN string
+	var annotationMap map[string]string
```
EronWright: In my opinion, the above variables would best be defined within the `for` loop.

Issue #226, opened 2019-07-09
Support passing of java options to bookkeeper from .yaml file


pbelgundi:
### Description
Currently values specified under "Options" in 'bookkeeper spec' in .yaml file for Pravega, are considered as bookkeeper server configuration options and there is no way to pass java options to the bookie process from a .yaml file.

### Importance
It is important to support this as we may need to tweak working of bookkeeper in some environments using java options, passed to the bookie process using the -D flag.

### Location
https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/bookie.go#L253

Pravega options:
Such options can be passed to the pravega image using "options" field in the spec. 
https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/pravega_controller.go#L164

Tristan1900:
Can I ask a question? What is the difference between bookkeeper server configuration options and java options(`-D`), I'm working on a related PR today and they look the same thing to me. Thanks! @pbelgundi

pbelgundi:
@Tristan1900 If you follow code links provided above, you will notice that "options" provided under "bookkeeper spec" are added to bookkeeper config map with the prefix "BK_". These are then added to "conf/bk_server.conf" file on the bookkeeper pod, as bookkeeper server options.
On the other hand, options provided to pravegaspec are prefixed with "-D" and provided as command line JVM options to controller and segment store processes.

As such, whatever K:V pairs are added under bookkeeper options, are not passed as JVM options to the bookkeeper process.

Tristan1900:
@pbelgundi Thanks for the explanation. What is exactly you want to pass using `-D`? 

From what I see, the `-D` is for java property and it can be passed using `-D` along with other non-standard(-X) and advanced(-XX) JVM options in the command line, but also can be passed using a config file, which is what bookkeeper is doing(using `conf/bk_server.conf`). Check [here](https://github.com/apache/bookkeeper/blob/master/bin/bookkeeper#L159). If you just want to pass java property to bookie similar to Pravega controller and segmentstore, then I think that has already been done. If you want to pass something else except java property using `-D`, please clarify what that is as I don't see such case right now.

The only missing part is to add the ability for users to pass non-standard(-X) and advanced(-XX) JVM options, which is not exactly what this issue describes.

pbelgundi:
@Tristan1900 Users may want to pass JVM options of any of these kinds: -D, -XX, or -X, to the bookkeeper server process.
[...]

PR #243, opened 2019-07-30
Issue 242: Preserve heap dump file

### LINKED ISSUE #242: Make sure heap dump file survives the pod restart
### Description

We are experiencing OOM error in segmentstore. When that happens the heap dump file is generated and stored locally in that pod and the pod is restarted instantly, so we lost the heap dump file for debugging. 

### Importance

Should have

### Location

Segmentstore and bookkeeper code

Use an emptyDir to store the heap dump file and specify the corresponding heap dump directory to that emptyDir.

Tristan1900:

### Change log description

This PR preserves the heap dump files during pod restart when OOM happens.

### Purpose of the change

Fix #242 

### What the code does

The heap dump will output to our specified directory, which is a mounted `emptyDIr`. Data in `emptyDIr` can survive the pod restart on the same node.

### How to verify it

Trigger an OOM, see if heap dump is in the specified directory after pod restarts.

### DIFF
--- a/pkg/controller/pravega/pravega_segmentstore.go
+++ b/pkg/controller/pravega/pravega_segmentstore.go
@@ -96,6 +96,10 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 						Name:      cacheVolumeName,
 						MountPath: cacheVolumeMountPoint,
 					},
+					{
+						Name:      heapDumpName,
+						MountPath: heapDumpDir,
+					},
 				},
 				Resources: *p.Spec.Pravega.SegmentStoreResources,
 				ReadinessProbe: &corev1.Probe{
@@ -129,6 +133,14 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 			},
 		},
 		Affinity: util.PodAntiAffinity("pravega-segmentstore", p.Name),
+		Volumes: []corev1.Volume{
+			{
+				Name: heapDumpName,
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
+			},
+		},
 	}
 
 	if p.Spec.Pravega.SegmentStoreServiceAccountName != "" {
@@ -145,9 +157,11 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
 	javaOpts := []string{
 		"-Xms1g",
+		"-XX:MaxDirectMemorySize=1g",
 		"-XX:+ExitOnOutOfMemoryError",
 		"-XX:+CrashOnOutOfMemoryError",
 		"-XX:+HeapDumpOnOutOfMemoryError",
+		"-XX:HeapDumpPath=" + heapDumpDir,
 		"-Dpravegaservice.clusterName=" + p.Name,
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -145,9 +157,11 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
 	javaOpts := []string{
 		"-Xms1g",
+		"-XX:MaxDirectMemorySize=1g",
```
andreipaduroiu: Can we make this configurable somehow? It would be very helpful for us if we could tweak this on the go via a manifest of some other way that doesn't include redeploying the operator.

Tristan1900: Thanks @andreipaduroiu , we should do that. This is also related to this [issue](https://github.com/pravega/pravega-operator/issues/226)

RaulGracia:
@Tristan1900 Maybe is not a big deal, but have you been able to test what happens in this situation?
1) A Segment Store crashes and write the dump with name like `java_pid1.dump`.
2) The Segment Store recovers and continues working.
3) The Segment Store crashes again and writes the dump, but the previous dump is still there and has the same file name.

In a previous experiment, I remember that I expected a newer dump file, but the old one was there (so, no file overwrite happened). Just want to mention that to confirm if you have observed the same behavior and, if so, if it is good enough like it is or if we want to keep more than one dump in case of subsequent crashes.

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: @Tristan1900 lets take up support for java options as part of a separate PR.

RaulGracia:
After a discussion with @pbelgundi, we found that making the JVM options/flags configurable to address @andreipaduroiu's comment is itself a piece of work that goes beyond the scope of the original issue. For this reason, the following could be done:
1) Leave in this PR only the changes related to the persistent storage of JVM heap dumps.
2) Create a new issue/PR targeted to make JVM options/flags configurable and set some reasonable defaults for both Bookkeeper and Pravega.

What do you think @Tristan1900?

pbelgundi:
Travis build is failing, perhaps some tests need fixing..

andreipaduroiu:
@Tristan1900 @RaulGracia I am perfectly fine with making another PR for the JVM options and memory. We need that to add more flexibility to our testing and deployment and it will be required if we implement direct-memory caching in the Segment Store. Let's make a new issue/PR and prioritize it for whenever the next release is.

pkg/controller/pravega/pravega_segmentstore.go
Tristan1900: agree

Tristan1900:
@RaulGracia Thanks for the comment. You are right about the dup dump files, I didn't consider that situation before. Will address that.
About raising another issue for options, I agree with that. Probably we can modify this [issue](https://github.com/pravega/pravega-operator/issues/226) to address this. What do you think? @pbelgundi  thanks.

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: > @RaulGracia Thanks for the comment. You are right about the dup dump files, I didn't consider that situation before. Currently, I'm not sure how to make heap dump files having different names. Is it ok to just maintain one?
> 
> About raising another issue for options, I agree with that. Probably we can modify this [issue](https://github.com/pravega/pravega-operator/issues/226) to address this. What do you think? @pbelgundi thanks.

Yes issue 226 is about passing JVM options to bookkeeper. For segment store and controller, it is possible to pass JVM options today, but those are one same set of options passed to both. That too needs to change. Lets address both these in PR for 226.

pbelgundi:
> @RaulGracia Thanks for the comment. You are right about the dup dump files, I didn't consider that situation before. Currently, I'm not sure how to make heap dump files having different names. Is it ok to just maintain one?
> 
> About raising another issue for options, I agree with that. Probably we can modify this [issue](https://github.com/pravega/pravega-operator/issues/226) to address this. What do you think? @pbelgundi thanks.

Yes, that would be good.

RaulGracia:
Yes @Tristan1900, maybe it is good to preserve just the first one. I will approve this PR, just recall to remove `-XX:MaxDirectMemorySize=1g`, as this change should be in another PR ok?

PR #255, opened 2019-08-30
Issue 153: Add Rollback support for Pravega Cluster

### LINKED ISSUE #153: Rollback when upgrade process fails
When the `version` field is updated on the `PravegaCluster` resource, the operator triggers a rolling upgrade process that updates BookKeeper, Segmentstore, and Controller pods until all of them are using the desired version. 

During the upgrade process, various errors can happen, from a version that does not exists to a failure in the configuration that prevents the pod from becoming ready. 

At this moment, when a failure occurs, the upgrade process is halted and requires manual intervention to restore the cluster's health. 

The operator should attempt to automatically restore the cluster health, rolling back to the original version. If the rollback fails, the process should halt and an alert should be sent.

pbelgundi:

### Change log description
- Support for manual rollback of Pravega cluster to the previous version when upgrade fails.
- Rollback can be done only to previous cluster version and only when cluster in 'Upgrade Failed' state.
- Added version history field to Pravega Cluster Status. This would store history of all versions through which PravegaCluster has been successfully upgraded so far. ( similar to kubectl rollout history ). The last value in this array would be same as Current Version.
- Improvements to Upgrade: When cluster upgrade is in progress, Cluster Status condition "Upgrading" is set to true and Reason and message values are set to indicate component currently being upgraded and number of replicas for which upgrade has completed against total replicas in component.
- Removed all occurrences of spec.version assignment from operator code.
- Added documentation for rollback.

### Purpose of the change
Fixes #153 

### How to verify it
1. Rollback is triggered  when Upgrade fails and user manually updates the spec.version field to the previous version.
2. When Rollback is in progress cluster condition `RollbackInProgress` set to Status=true.
3. Rollback of Pravega cluster to the previous version completes successfully for all Pravega Components.
This is indicated by all pravega replicas moving back to the older version and re-starting successfully.
Tested rollback after a failing upgrade in each component - bookkeeper, controller and segmentstore  and in all 3 cases the cluster was successfully restored back to the previous version.
3. Incase of Rollback Failure, Cluster Status is set to Condition `Error`  with Status=true and Reason = `RollbackFailed` and manual intervention would be needed to recover from this state.


pkg/controller/pravegacluster/upgrade.go
```diff
@@ -108,9 +115,60 @@ func (r *ReconcilePravegaCluster) clearUpgradeStatus(p *pravegav1alpha1.PravegaC
...
 
-func (r *ReconcilePravegaCluster) syncComponentsVersion(p *pravegav1alpha1.PravegaCluster) (err error) {
-	var synced bool
+func (r *ReconcilePravegaCluster) rollbackClusterVersion(p *pravegav1alpha1.PravegaCluster, version string) (err error) {
+	_, rollbackCondition := p.Status.GetClusterCondition(pravegav1alpha1.ClusterConditionRollback)
+	if rollbackCondition == nil {
+		// We're in the first iteration for Rollback
+		// Add Rollback Condition to Cluster Status
+		p.Status.SetRollbackConditionTrue()
+		p.Spec.Version = version
```
EronWright: I am leery of changing the `PravegaCluster` spec; it is the user's goal, and is declaratively managed by configuration files near the installer.   The status can change to show that the upgrade was unsuccessful, but I think the user should trigger rollback by updating the goal.

pbelgundi: Agree that cluster `spec` represents the `desired` state and a controller(operator) should not change that value, since this is also the behavior exhibited by kubernetes when handling upgrades to deployments and statefulsets.

However, implementing automatic rollback from within the operator would require updating the cluster spec.
Automated rollback is what most customers would want in case of an upgrade failure as it removes the need for manual intervention and also makes sure the cluster stays in inconsistent state ( few pods will old and few with new version) for the shortest time.

A way to support both could be to have the operator do a manually triggered rollback  and automate invocation of this rollback when upgrade fails through some external means.
Please let me know your thoughts on this.

pbelgundi: I tried implementing Automated Rollback without changing `spec.Version` and rollback completed successfully, however post completion of rollback it triggers upgrade again, as the trigger for upgrade is `p.Spec.Version != p.Status.CurrentVersion`.

pkg/apis/pravega/v1alpha1/status.go
```diff
@@ -165,3 +196,30 @@ func (ps *ClusterStatus) setClusterCondition(newCondition ClusterCondition) {
 
 	ps.Conditions[position] = *existingCondition
 }
+
+func (ps *ClusterStatus) AddToVersionHistory(version string) {
+	lastIndex := len(ps.VersionHistory) - 1
```
Tristan1900: Could the length of `versionHistroy` be 0 here?

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -108,9 +115,60 @@ func (r *ReconcilePravegaCluster) clearUpgradeStatus(p *pravegav1alpha1.PravegaC
...
+		p.Status.TargetVersion = p.Spec.Version
+		updateErr := r.client.Status().Update(context.TODO(), p)
+		if updateErr != nil {
+			log.Printf("Error updating cluster: %v", updateErr.Error())
+			return fmt.Errorf("Error updating cluster status. %v", updateErr)
+		}
+	}
+
+	syncCompleted, err := r.syncComponentsVersion(p)
```
Tristan1900: For the rollback, I'm not sure the execution sequence should be the same as the upgrade. Say we have an upgraded bookkeeper client that has some new features in the updated segmentstore. If we roll back the bookkeeper server first, is it possible that the segmentstore has errors due to its bookkeeper client code uses the new feature against an old version bookkeeper server? It occurs to me because in the bookkeeper [doc](https://bookkeeper.apache.org/docs/4.7.2/admin/upgrade/), it says usually we need to upgrade server before client. So for rollback, the sequence probably should be reversed somehow?

pkg/controller/pravegacluster/upgrade.go
pbelgundi: Removed all occurrences of assignment to `p.Spec.Version` in operator code.
Implemented a manual rollback trigger when upgrade fails

pkg/controller/pravegacluster/upgrade.go
pbelgundi: Yes. That sound like a valid point. Will consider reversing the order

bourgeoisor:
Thanks for modifying the flow to cater mostly to manual upgrade, Prajakta! Some of these tweaks are also really helpful, like the enhanced statuses. I have gone through the diff and this approach seems good to me. That being said, I would like someone else to approve in my stead, mainly due to my lack of deep understanding of the Pravega codebase and practices.

pbelgundi:
> Thanks for modifying the flow to cater mostly to manual upgrade, Prajakta! Some of these tweaks are also really helpful, like the enhanced statuses. I have gone through the diff and this approach seems good to me. That being said, I would like someone else to approve in my stead, mainly due to my lack of deep understanding of the Pravega codebase and practices.

@bourgeoisor, thanks for your comments. 
I have added @fpj as a reviewer.

pkg/controller/pravegacluster/upgrade.go
pbelgundi: Changed Rollback to updates components in reverse order as upgrade.

pkg/controller/pravegacluster/upgrade_test.go
```diff
@@ -170,7 +170,7 @@ var _ = Describe("Pravega Cluster", func() {
 
 				It("should set upgrade condition reason to UpgradingBookkeeperReason and message to 0", func() {
 					_, upgradeCondition := foundPravega.Status.GetClusterCondition(pravegav1alpha1.ClusterConditionUpgrading)
-					Ω(upgradeCondition.Reason).Should(Equal(pravegav1alpha1.UpgradingBookkeeperReason))
+					Ω(upgradeCondition.Reason).Should(Equal(pravegav1alpha1.UpdatingBookkeeperReason))
```
fpj: The only changes to testing are naming changes, can we add test cases to validate functionality?

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -151,6 +158,7 @@ func (r *ReconcilePravegaCluster) run(p *pravegav1alpha1.PravegaCluster) (err er
 }
 
 func (r *ReconcilePravegaCluster) deployCluster(p *pravegav1alpha1.PravegaCluster) (err error) {
+
```
fpj: These new lines after the function signature don't seem to be added consistently.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -111,10 +121,61 @@ func (r *ReconcilePravegaCluster) clearUpgradeStatus(p *pravegav1alpha1.PravegaC
...
-	var synced bool
+func (r *ReconcilePravegaCluster) rollbackClusterVersion(p *pravegav1alpha1.PravegaCluster, version string) (err error) {
+	_, rollbackCondition := p.Status.GetClusterCondition(pravegav1alpha1.ClusterConditionRollback)
+	if rollbackCondition == nil || rollbackCondition.Status != corev1.ConditionTrue {
+		// We're in the first iteration for Rollback
+		// Add Rollback Condition to Cluster Status
+		log.Printf("Updating Target Version to  %v", version)
+		p.Status.TargetVersion = version
+		p.Status.SetRollbackConditionTrue("", "")
+		updateErr := r.client.Status().Update(context.TODO(), p)
```
fpj: What's `context.TODO()`?

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -111,10 +121,61 @@ func (r *ReconcilePravegaCluster) clearUpgradeStatus(p *pravegav1alpha1.PravegaC
...
+		if updateErr != nil {
+			p.Status.SetRollbackConditionFalse()
+			log.Printf("Error updating cluster: %v", updateErr.Error())
+			return fmt.Errorf("Error updating cluster status. %v", updateErr)
+		}
+	}
+
+	syncCompleted, err := r.syncComponentsVersion(p)
+	if err != nil {
+		// error rolling back, set appropriate status and ask for manual intervention
```
fpj: You're not capitalizing the first letter of comments consistently.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -111,10 +121,61 @@ func (r *ReconcilePravegaCluster) clearUpgradeStatus(p *pravegav1alpha1.PravegaC
 	return nil
 }
 
-func (r *ReconcilePravegaCluster) syncComponentsVersion(p *pravegav1alpha1.PravegaCluster) (err error) {
-	var synced bool
+func (r *ReconcilePravegaCluster) rollbackClusterVersion(p *pravegav1alpha1.PravegaCluster, version string) (err error) {
```
fpj: Say the operator crashes and restarts in the middle of a triggered rollback. What happens in this case? Are we able to pick up from where the previous instance left off?

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -111,10 +121,61 @@ func (r *ReconcilePravegaCluster) clearUpgradeStatus(p *pravegav1alpha1.PravegaC
...
+		p.Status.SetRollbackConditionTrue("", "")
+		updateErr := r.client.Status().Update(context.TODO(), p)
+		if updateErr != nil {
+			p.Status.SetRollbackConditionFalse()
+			log.Printf("Error updating cluster: %v", updateErr.Error())
+			return fmt.Errorf("Error updating cluster status. %v", updateErr)
+		}
+	}
+
+	syncCompleted, err := r.syncComponentsVersion(p)
```
fpj: For my understanding, does this call block until the update of all components complete?

pkg/controller/pravegacluster/upgrade.go
pbelgundi: Yes. We would be able to pick up from where we left off last.

pkg/controller/pravegacluster/upgrade.go
pbelgundi: Nope. It updates the spec of one component and then on every reconciliation loop checks how many replicas are updated for that component. When all are updated, we move to next component till all components are upgraded.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -339,31 +415,25 @@ func (r *ReconcilePravegaCluster) syncBookkeeperVersion(p *pravegav1alpha1.Prave
...
 	// Pod template already updated
-
 	log.Printf("statefulset (%s) status: %d updated, %d ready, %d target", sts.Name,
 		sts.Status.UpdatedReplicas, sts.Status.ReadyReplicas, sts.Status.Replicas)
 
 	// Check whether the upgrade is in progress or has completed
 	if sts.Status.UpdatedReplicas == sts.Status.Replicas &&
 		sts.Status.UpdatedReplicas == sts.Status.ReadyReplicas {
 		// StatefulSet upgrade completed
-		// TODO: wait until there is no under replicated ledger
```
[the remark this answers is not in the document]
pbelgundi: Because there us nothing to be done for under-replicated ledgers here. Autorecovery takes care of that.

pkg/controller/pravegacluster/upgrade_test.go
pbelgundi: I have added these today. Will add more tomorrow.

doc/rollback-cluster.md
```diff
@@ -0,0 +1,101 @@
...
+
+Manual intervention would be needed for resolving this.
+
+
+
+
+
+
+
+## Pending tasks
```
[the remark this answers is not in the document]
pbelgundi: fixed

pkg/apis/pravega/v1alpha1/status.go
```diff
@@ -117,6 +128,16 @@ func (ps *ClusterStatus) SetUpgradingConditionFalse() {
 	ps.setClusterCondition(*c)
 }
 
+/*
```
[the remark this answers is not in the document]
pbelgundi: done

pkg/controller/pravegacluster/pravegacluster_controller.go
pbelgundi: removed

pkg/controller/pravegacluster/upgrade.go
pbelgundi: context.TODO() creates an empty context.
https://blog.golang.org/context

doc/rollback-cluster.md
```diff
@@ -0,0 +1,101 @@
...
+If Rollback completes successfully, cluster state goes back to `PodsReady` which would mean the cluster is now in a stable state.
+If Rollback Fails, the cluster would move to state `RollbackFailed` indicated by this cluster condition:
+```
+ClusterConditionType: Error
+Status: True
+Reason: RollbackFailed
+Message: <Details of exception/cause of failure>
+```
+
+Manual intervention would be needed for resolving this.
```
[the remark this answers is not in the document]
pbelgundi: Manual steps can vary depending on root cause of failure.

pkg/controller/pravegacluster/upgrade.go
fpj: The documentation says:

> TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

https://golang.org/pkg/context/#TODO

Could you justify why you are using it a good number of times in this PR?

pkg/controller/pravegacluster/upgrade.go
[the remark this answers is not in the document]
fpj: Auto-replication in BK does take care of replicating data, but we don't actually know when that happens for any given ledger. Consequently, a ledger can be under-replicated for arbitrarily long.

I believe this comment is saying that, before declaring an upgrade complete, we should make sure that all under-replicated ledgers have been properly replicated.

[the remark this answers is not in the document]
pbelgundi: > Auto-replication in BK does take care of replicating data, but we don't actually know when that happens for any given ledger. Consequently, a ledger can be under-replicated for arbitrarily long.
> 
> I believe this comment is saying that, before declaring an upgrade complete, we should make sure that all under-replicated ledgers have been properly replicated.

Auto replication kicks in if a bookie is down for more than a certain configurable time interval, and makes sure under-replicated ledgers are dealt with. I have documented the details of this in #235.   
We would need to tune auto-recovery configuration parameters based on how long we want to tolerate data loss or how soon we need the recovery to start.
Nevertheless, this not an upgrade problem but a general DR issue, where a certain server or set of servers could be down for an arbitrary period of time (due to upgrade or otherwise) and data from the same needs to be recovered.

pkg/controller/pravegacluster/upgrade.go
pbelgundi: All operator code I have seen uses "context.TODO()" in every client call. 
While it does seem to work, I agree it could be suboptimal. Need to spend some time understanding this, its impact and alternatives.

pkg/controller/pravegacluster/upgrade.go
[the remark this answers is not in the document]
fpj: >  this not an upgrade problem but a general DR issue

there is a difference between data being unavailable or under-replicated due to say a crash and the data being unavailable or under-replicated due to a planned task like an upgrade. maybe this is not for this PR, but I don't think it is acceptable that post upgrade the cluster is in a worse state (e.g., under-replicated data) compared to what it was prior to upgrade starting.

pkg/controller/pravegacluster/upgrade.go
fpj: could create an issue and reference it here, please?

doc/rollback-cluster.md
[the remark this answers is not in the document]
fpj: It is fine that you can't describe it exhaustively, but could you give at least one example?

pkg/controller/pravegacluster/upgrade.go
pbelgundi: https://github.com/pravega/pravega-operator/issues/266

pkg/controller/pravegacluster/upgrade.go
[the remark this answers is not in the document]
pbelgundi: Note that under-replication introduced by upgrade is temporary (till the node being upgraded comes back - about 2-3 mins ) and should be acceptable.
Bookkeeper documentation infact suggests switching off Autorecovery during maintenance tasks like upgrade, as the under-replication is only temporary and should not cause auto-recovery to trigger unnecessarily:
https://bookkeeper.apache.org/docs/4.7.3/admin/autorecovery/
```
You can disable AutoRecovery at any time, for example during maintenance. Disabling AutoRecovery ensures that bookies’ data isn’t unnecessarily rereplicated when the bookie is only taken down for a short period of time, for example when the bookie is being updated or the configuration if being changed.```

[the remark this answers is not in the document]
pbelgundi: I don't see how a cluster could be in a worse state post upgrade. 
Data replication is really not related to upgrade.
The right auto-recovery configuration would make sure under-replicated data is replicated every 'x' mins, irrespective of whether a node is down due to upgrade or other issues.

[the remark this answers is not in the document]
pbelgundi: In terms of limiting the damage upgrade can cause due to nodes being down, I have capped the max time a node can take to come back up to 10 mins, failing which we fail the entire upgrade process and rollback should bring the cluster back to a normal state.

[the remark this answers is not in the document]
fpj: I keep forgetting that we don't wipe out the bookie data. Perhaps we should consider turning the auto-replication off during an upgrade as you mention, @pbelgundi . We should create an issue so that we can assess whether it is necessary.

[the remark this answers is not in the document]
pbelgundi: Issue for tracking auto-recovery enable/disable:
https://github.com/pravega/pravega-operator/issues/267

doc/rollback-cluster.md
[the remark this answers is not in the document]
pbelgundi: I have updated the documentation to reflect what can be done to recover from a `RollbackFailed` state.

PR #264, opened 2019-09-17
Issue 262: Adding Helm Test for Pravega Cluster

### LINKED ISSUE #262: Pravega Cluster health check
### Description
This is to implement `helm test` on the Pravega Cluster so that post an upgrade we can determine whether the upgrade was successful or not by simply running a `helm test <release-name>`.

SrishT:

### Change log description
Implementing helm test on the Pravega Cluster so that post an upgrade we can determine whether the upgrade was successful or not by simply running `helm test <release-name>`.

### Purpose of the change
Fixes #262 

### What the code does
The helm test simply checks whether `PodsReady` status field of the Pravega Cluster is set to true or not which can be used to determine whether the Pravega Cluster is ready or not.

### How to verify it
Run `helm test <release-name>`
If the PravegaCluster upgrade was successful and all pods are in Ready state then the test should pass, else it should fail.

### DIFF
--- a/charts/pravega/templates/tests/clusterrole.yaml
+++ b/charts/pravega/templates/tests/clusterrole.yaml
@@ -1,0 +1,23 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: "{{.Release.Name}}-pravega-test"
+rules:
+- apiGroups:
+  - pravega.pravega.io
+  resources:
+  - "*"
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - "*"
+  verbs:
+  - '*'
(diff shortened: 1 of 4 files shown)


charts/pravega/templates/tests/clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
+kind: ClusterRole
```
pbelgundi: Since these templates for test, I would prefer these be added under "pravega-operator/charts/test" instead of the current location.

SrishT: By definition, a test in a helm chart lives under the `templates/` directory and is a pod definition that specifies a container with a given command to run.
Reference : https://helm.sh/docs/developing_charts/

charts/pravega/templates/tests/clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: "{{.Release.Name}}-pravega-test"
+rules:
+- apiGroups:
+  - pravega.pravega.io
+  resources:
+  - "*"
+  verbs:
+  - "*"
```
tewit: Does the test pod need to have all access? It seems the condition check is a read-only operation.

SrishT: Fixed this

Issue #245, opened 2019-08-06
Add support for configuring external access separately for Controller and Segment Store


pbelgundi:
### Description
Controller and Segment store external services may require to be configured differently. However, currently there is only 1 ExternalAccess field in PravegaSpec that is used to get values for services of both these components. It would help to refactor that to have separate fields for controller and segment store services.

### Importance
Needed for Pravega-185

### Suggestions for an improvement
This could be part of a greater CRD refactoring. See issue #85

pbelgundi:
This is fixed with above PRs.

PR #268, opened 2019-09-19
Issue:244,255 Support custom annotations and separate service types for external access for Controller and Segment Store

### LINKED ISSUE #244: Pravega Operator should support addition of custom annotations on Controller and Segment Store
### Description
At deployment time it is impossible for operator to know ofall annotations that users may need add to services for their reference or use later. As such, operator should allow annotations to be specified via manifest that can be added to Controller Service and Segment store services.

### Importance
Needed for Pravega-183

Currently there is no way to specify separate annotations for Controller and Segment Store. 
See related issue: #85

pbelgundi:

### Change log description
Enables different service types to be provided for Controller and SegmentStore for external access, while maintaining backward compatibility with existing Cluster Spec.
Adds support for specifying custom annotations on controller and segment store external services.

### Purpose of the change
Fixes #244, #245

### What the code does
Code uses external access service type from fields `ControllerExternalServiceType` for controller and `SegmentStoreExternalServiceType` for SegmentStore under `spec.pravega`, if these are specified.
If these are not specified in the manifest, service type specified under `spec.ExternalAccess.Type` is used for both controller and segment store services.
If both values are not provided and external access is enabled, defaultServiceType (LoadBalancer) is used.
Also two new fields added for Annotations under `spec.Pravega`
```
ControllerServiceAnnotations
SegmentStoreServiceAnnotations
```

### How to verify it
All Unit Tests and E2E Tests should pass after this change.

### DIFF
--- a/doc/external-access.md
+++ b/doc/external-access.md
@@ -170,3 +170,71 @@ metadata:
 . . .
 
 ```
+
+3. Overriding Service Type for Controller and SegmentStore external services.
+When External Access is enabled, if ServiceType for Controller and SegmentStore is same, it can be specified under `externalAccess` > `type` as shown earlier.
+However, if we need to specify different service types for Controller and SegmentStore services, this can be done by adding `controllerExtServiceType` or `segmentStoreExtServiceType` field for Controller and SegmentStore respectively, under the PravegaSpec.
+When specified, these values will override the value against `externalAccess`, `type` field, if present.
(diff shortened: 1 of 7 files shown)


pbelgundi:
> I see API changes in this PR but I don't see deepcopy files change.

Updated the deepcopy file and also added tests for this change.

PR #278, opened 2019-09-26
Issue 273 srs events with transform

ravibeta:
### Change log description
This is a sample code to specify K8s event specific for use with diagnostics. It differs from the existing event in the use of involved object and a few other fields.

### Purpose of the change

This sample code can be merged as-is or used to enhance the existing event to have all the necessary information 

### What the code does

raises a K8s event specific to diagnostics whenever the corresponding custom K8s event is issued.

### How to verify it

the get events command will have this event corresponding to this namespace.

### DIFF
--- a/pkg/util/k8sutil.go
+++ b/pkg/util/k8sutil.go
@@ -176,7 +176,7 @@ func NewK8sEvent(name string, p *v1alpha1.PravegaCluster, reason string, message
 	labels[SRSEvent] = trueString
 	event := corev1.Event{
 		Count: 1,
-		InvolvedObject: v1.ObjectReference{
+		InvolvedObject: corev1.ObjectReference{
 			APIVersion: "apps/v1",
 			Kind:       "Application",
 			Name:       name,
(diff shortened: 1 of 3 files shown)


pkg/util/k8sutil.go
```diff
@@ -121,3 +144,64 @@ func IsPodFaulty(pod *corev1.Pod) (bool, error) {
...
+		ReportingInstance:   os.Getenv("POD_NAME"),
+	}
+	return &event
+}
+
+func NewK8sEvent(name string, p *v1alpha1.PravegaCluster, reason string, message string, eventType string) *corev1.Event {
+	var labels = make(map[string]string)
+	labels[KAHMCustomLabel] = trueString
+	labels[SymptomIDLabel] = string(CriticalEvent)
+	labels[SRSEvent] = trueString
```
maddisondavid: Pravega Operator should be agnostic to any monitoring system

maddisondavid:
Also note that some of the commits are failing DCO checking because they are not signed.

Issue #294, opened 2019-10-30
Create an uber helm chart.


pbelgundi:
### Description
Create an uber helm chart or script that deploys all dependencies (Operators, Zookeeper, Bookkeeper) and Pravega components in the right order.

### Importance
 must-have

### Location
Under charts folder in pravega-operator code

### Suggestions for an improvement
Have a single helm chart or script that deploys not just pravega components but also other supporting software like operator, zookeeper, zookeeper-operator, tier2 etc...in the right order. This chart should also ensure that the correct/compatible versions of dependencies are used when deploying Pravega

PR #319, opened 2020-01-24
Issue 315: Bookkeeper Cluster should not be managed by Pravega Operator

### LINKED ISSUE #315: Remove Bookkeeper from Pravega Operator
### Description
Currently, the operator couples the deployment of Bookkeeper and Pravega. However, in some scenarios, a user may already have a Bookkeeper cluster ready and independently managed. In this case, the operator should skip the deployment of Bookkeeper and just ask for the endpoint of the external Bookkeeper cluster to use.

### Importance
Must-have feature.

### Location
Bookkeeper deployment options.

Similarly to how Zookeeper is configured, we can add a new `bookkeeperUrl` field to point to an existing Bookkeeper cluster. We may need to make sure that wither this field or the current Bookkeeper manifest configuration to deploy Bookkeeper are filled in the manifest.

pbelgundi:
### Change log description
Code changes for removing BK from Pravega Spec

### Purpose of the change
Fixes #315

### How to verify it
PravegaOperator should take BookkeeperUri as input in the Spec similar to ZookeeperUri. If not specified, default value will be used. Post this fix, Pravega-Operator will not be responsible for Bookkeeper cluster deployment and management and Bookkeeper cluster should be setup independently prior to setting up Pravega Cluster using this operator.


charts/pravega/README.md
```diff
@@ -42,13 +42,7 @@ The following table lists the configurable parameters of the Pravega chart and t
...
 | `externalAccess.enabled` | Enable Pravega external access | `false` |
 | `externalAccess.type` | Pravega external access type | `LoadBalancer` |
-| `bookkeeper.image.repository` | Image repo for Bookkeeper image | `pravega/bookkeeper` |
-| `bookkeeper.replicas` | Replicas for Bookkeeper | `3` |
-| `bookkeeper.storage.className` | Storage class for Bookkeeper volumes | `standard` |
-| `bookkeeper.storage.ledgerVolumeRequest` | Request storage for ledgerVolume | `10Gi` |
-| `bookkeeper.storage.journalVolumeRequest` | Request storage for journalVolume | `10Gi` |
-| `bookkeeper.storage.indexVolumeRequest` | Request storage for indexVolume | `10Gi` |
-| `bookkeeper.autoRecovery`| Enable Bookkeeper autoRecovery | `true` |
+| `bookkeeperUri` | URL for Bookkeeper Servers | `bookkeeper-client:3181` |
```
RaulGracia: I think that we cannot merge this PR until the Bookkeeper operator is developed, tested and released (initial release). Is that reasoning correct? Otherwise, users won't be able to deploy Pravega in the meantime.

RaulGracia: Also, the documentation needs to point to the new Bookkeeper operator as a pre-requisite prior to deploy this Operator and Pravega, right?

RaulGracia: Actually, in the meantime I think it makes sense to have this feature: https://github.com/pravega/pravega-operator/tree/pravega-without-bk

test/e2e/scale_test.go
```diff
@@ -46,7 +46,6 @@ func testScaleCluster(t *testing.T) {
 	g.Expect(err).NotTo(HaveOccurred())
 
 	// Scale up Pravega cluster, increase bookies and segment store size by 1
-	pravega.Spec.Bookkeeper.Replicas = 4
```
RaulGracia: How are we going to perform test when this change is merged? Are we going to "mock" Bookies?

charts/pravega/README.md
pbelgundi: Making bookkeeper "optional" complicates a lot of things with p-operator which is why this option of removing it completely from p-operator was choosen. I would not recommend it at all. Once the BK PR is merged, I don't see why we should be blocked on merging this PR.

pbelgundi: Sure. Will add this. While we can't mandate that we can recommend it.

test/e2e/scale_test.go
pbelgundi: For unit tests there is BK is not needed. For E2E tests we would do the same things we do for ZK.

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -196,16 +196,19 @@ func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
...
-	for i := int32(0); i < util.Min(3, p.Spec.Bookkeeper.Replicas); i++ {
-		waitFor = append(waitFor,
-			fmt.Sprintf("%s-%d.%s.%s:3181",
-				util.StatefulSetNameForBookie(p.Name),
-				i,
-				util.HeadlessServiceNameForBookie(p.Name),
-				p.Namespace))
-	}
-	configData["WAIT_FOR"] = strings.Join(waitFor, ",")
+	/*
```
Tristan1900: Perhaps we could delete these comments?

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -259,6 +248,7 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
 	return nil
 }
 
+/*
```
Tristan1900: same here

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -314,6 +299,7 @@ func (r *ReconcilePravegaCluster) syncClusterSize(p *pravegav1alpha1.PravegaClus
...
```
Tristan1900: same here

pkg/controller/pravegacluster/upgrade_test.go
```diff
@@ -150,50 +150,17 @@ var _ = Describe("Pravega Cluster Version Sync", func() {
...
-					Ω(upgradeCondition.Reason).Should(Equal(pravegav1alpha1.UpdatingBookkeeperReason))
-					Ω(upgradeCondition.Message).Should(Equal("0"))
-				})
-			})
-
 			Context("Upgrade Segmentstore", func() {
 				var (
 					foundPravega *v1alpha1.PravegaCluster
-					sts          *appsv1.StatefulSet
+					//sts          *appsv1.StatefulSet
```
Tristan1900: same here

pkg/controller/pravegacluster/upgrade_test.go
```diff
@@ -150,50 +150,17 @@ var _ = Describe("Pravega Cluster Version Sync", func() {
...
-					sts = &appsv1.StatefulSet{}
-					name := util.StatefulSetNameForBookie(p.Name)
-					_ = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, sts)
-					targetImage, _ := util.BookkeeperTargetImage(foundPravega)
-					sts.Spec.Template.Spec.Containers[0].Image = targetImage
-					r.client.Update(context.TODO(), sts)
-
 					_, _ = r.Reconcile(req)
-					_ = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, sts)
+					//_ = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, sts)
```
Tristan1900: here as well

pkg/webhook/webhook.go
```diff
@@ -98,9 +98,6 @@ func (pwh *pravegaWebhookHandler) mutatePravegaVersion(ctx context.Context, p *p
 	if p.Spec.Pravega != nil && p.Spec.Pravega.Image != nil && p.Spec.Pravega.Image.Tag != "" {
 		p.Spec.Pravega.Image.Tag = ""
 	}
-	if p.Spec.Bookkeeper != nil && p.Spec.Bookkeeper.Image != nil && p.Spec.Bookkeeper.Image.Tag != "" {
```
Tristan1900: I guess we need to find a new way to ensure that Pravega and Bookkeeper are using the compatible versions.

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
```diff
@@ -19,12 +19,15 @@ const (
 	// DefaultZookeeperUri is the default ZooKeeper URI in the form of "hostname:port"
 	DefaultZookeeperUri = "zk-client:2181"
 
+	// DefaultBookkeeperUri is the default ZooKeeper URI in the form of "hostname:port"
```
Tristan1900: Just wondering if we can have a Uri like `DefaultZookeeperUri = "zk-client:2181"`? If we have like 6 Bookies then user need to manually type a bunch of words. 
Maybe we can have something like 
```
Bookkeeper:
  name: 
  replicas:
```
and we generate the name for users. What do you think?

pbelgundi: Good point. However, we can't **assume** that the BK deployment used with Pravega will be a statefulset with a headless service and that users will use the headless service URLs for Bookkeeper here. One option could be to create a client service for the  Bookkeeper Cluster similar to ZK and expect them to provide the URL for the same.

pkg/controller/pravega/pravega_segmentstore.go
pbelgundi: Sure.

pkg/controller/pravegacluster/pravegacluster_controller.go
pbelgundi: ok

pkg/controller/pravegacluster/pravegacluster_controller.go
pbelgundi: Ok

pkg/webhook/webhook.go
pbelgundi: That would be managed using an uber helm chart that will deploy all pravega dependencies and components. See: https://github.com/pravega/pravega-operator/issues/294

pkg/controller/pravegacluster/upgrade_test.go
pbelgundi: done.

pkg/controller/pravegacluster/upgrade_test.go
pbelgundi: done

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
pbelgundi: Headless services can also be referenced using the service URL instead for hostnames, like: pravega-bk-headless.default.svc.cluster.local
See: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services

README.md
```diff
@@ -39,6 +39,7 @@ The Pravega Operator manages Pravega clusters deployed to Kubernetes and automat
 - Kubernetes 1.9+
 - Helm 2.10+
 - An existing Apache Zookeeper 3.5 cluster. This can be easily deployed using our [Zookeeper operator](https://github.com/pravega/zookeeper-operator)
+- An existing Apache Bookkeeper 4.7.3 cluster. This can be easily deployed using our [BookKeeper Operator](https://github.com/pravega/bookkeeper-operator)
```
RaulGracia: This version should be updated, right? We now moved to Bookkeeper 4.9.2.

pbelgundi: fixed

PR #328, opened 2020-02-27
Issue 314: Go version changed to 1.13 so as to switch use to go modules

### LINKED ISSUE #314: Upgrade GO Version and switch to using GO Modules
### Description

The Go Version currently being used by operator is pretty old and it would be good to upgrade to a newer version like go1.13. Also need to switch to using Go Modules to get over the limitation of having to create the project under $GOPATH.

### Importance

_(Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have))_

### Location

_(Where is the piece of code, package, or document affected by this issue?)_

_(How do you suggest to fix or proceed with this issue?)_

Prabhaker24:
# Purpose of the change
Fixes #314

# what the code does:
previously we were not using go modules because of which we had a restriction of having the code in the go path for it to work with these changes we will be able to even run our code without it being in the go path

# How to verify it :
1) All the make commands in the make file are working as expected
2) Checkout and build code outside $GOPATH should work.
3)  All basic operator functionality - Cluster Creation, Deletion, Scaling-up & Down, Upgrades and 
    Rollback work as expected.

**Note**- I have verified all the above changes


Makefile
```diff
@@ -16,14 +16,11 @@ TEST_IMAGE=$(REPO)-testimages:$(VERSION)
 GOOS=linux
 GOARCH=amd64
 
-.PHONY: all dep build check clean test
+.PHONY: all  build check clean test
```
pbelgundi: Please update development documentation under `doc/development.md` to reflect right way to pull dependencies after moving to go modules.

pbelgundi: Gopkg.toml is needed only by `dep` please remove that file.

Makefile
```diff
@@ -16,14 +16,11 @@ TEST_IMAGE=$(REPO)-testimages:$(VERSION)
...
 
-all: check test build
+all: check build 
 
-dep:
-	dep ensure -v
-
-build: dep build-go build-image
+build: build-go build-image
```
pbelgundi: `make dep` is removed as we no longer rely on `dep` to pull our dependencies. But we perhaps need to add `go mod tidy` here. Else in a new environment how does the user know how to pull dependencies?

pbelgundi:
Please update the PR with details about how this change was tested.
The following needs to be tested:

1) All "make" commands( except `dep`) should work as earlier after this change.
2) We should be able to checkout and build code outside $GOPATH.
3) Since there has been a change to operator base image, please build an operator image and using this image that verify that all basic operator functionality - Cluster Creation, Deletion, Scaling up & Down, Upgrades and Rollback work as expected after this change.

go.mod
```diff
@@ -0,0 +1,98 @@
+module github.com/pravega/pravega-operator
```
pbelgundi: Since this file is added, files  `Gopkg.lock` and `Gopkg.toml` can be removed.

Makefile
pbelgundi: We perhaps also need to set `export GO111MODULE=on`.... please check if its needed and if yes, create a new make target that will do all necessary steps for pulling dependencies using gomodules. This target should replace the earlier "dep" target everywhere in the Makefile.
https://github.com/golang/go/wiki/Modules#how-to-define-a-module

pbelgundi:
A good reference for migrating from dep to modules:
https://medium.com/@elliotchance/migrating-projects-from-dep-to-go-modules-cc6bde71610b

Dockerfile
```diff
@@ -7,35 +7,42 @@
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
-FROM golang:1.10.1-alpine3.7 as go-builder
+ARG GO_VERSION=1.13.4
```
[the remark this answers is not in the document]
Prabhaker24: I have changed the version to 1.13.8

Makefile
```diff
@@ -16,14 +16,11 @@ TEST_IMAGE=$(REPO)-testimages:$(VERSION)
 GOOS=linux
 GOARCH=amd64
 
-.PHONY: all dep build check clean test
+.PHONY: all  build check clean test
 
-all: check test build
+all: check build
```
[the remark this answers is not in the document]
Prabhaker24: removed it.

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: required
 language: go
 go_import_path: github.com/pravega/pravega-operator
 go:
-  - '1.11'
+  - '1.13'
```
pbelgundi: This should be `1.13.8`

Makefile
Prabhaker24: we don't need to do make dep as while running go build dependencies will be pulled based on go.mod file

Prabhaker24: We don't need export GO111MODULE=on as we are using go version 1.13 which by default will use the go.mod file if present this property value is needed for version below 1.13 that to when the project is inside the GOPATH to make use of go.mod file while downloading dependencies

go.mod
Prabhaker24: removed both of the files

Makefile
Prabhaker24: I have updated the document.

Prabhaker24: I have updated the document.

doc/development.md
```diff
@@ -2,7 +2,7 @@
 
 Requirements:
   - Go 1.10+
```
pbelgundi: This needs to be changed to Go 1.13+

doc/development.md
```diff
@@ -24,13 +24,13 @@ gvm listall
 ```
 
 Install a new go version:
-``` 
-gvm install go1.4 -B 
+```
+gvm install go1.4 -B
 gvm use go1.4
 gvm install go1.11 --binary
```
pbelgundi: Change to go `1.13.8`

doc/development.md
Prabhaker24: We can use go version above 11 with go modules just that we need to set the GO111MODULE=on in case we are inside the go path to use go modules I have mentioned that in the document as well

doc/development.md
Prabhaker24: I have changed it

doc/development.md
```diff
@@ -39,23 +39,23 @@ should display something like `/home/<userdir>/.gvm/pkgsets/go1.11/global`
...
+go get github.com/pravega/pravega-operator
 ```
 This should clone operator code under `$GOPATH/src/github.com/pravega/pravega-operator`
 
-### Install dep
-```
-curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-```
-This may prompt for creation of `GOBIN` directory. Please create it and then issue this command again.
+Set the following system property to use go modules when the version of go is below 1.13 and the operator code is cloned inside the GOPATH
```
pbelgundi: This text can be removed since our minimum version is 1.13+
```
Set the following system property to use go modules when the version of go is below 1.13 and the operator code is cloned inside the GOPATH
```	
curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh	`GO111MODULE=on`
```

Prabhaker24: Done

PR #333, opened 2020-03-05
Issue 332:  Allow ingestion of user defined variables into the SegmentStore

### LINKED ISSUE #332: Pravega Operator should allow ingestion of user defined env vars and secrets in Pravega Components
### Description
We should have the provision of environmental variables into the SegmentStore container, like
```
VCENTER_USER: ...
VCENTER_PASSWORD: ...
VM_NAME_FILTER_PREFIX: ...
VCENTER_IP: ...
```
The Pravega Operator should enable this by allowing users to create a custom config map and associate the same with segmentstore in pravega spec

### Importance
must-have

Going forward we might need to provide the username and password as a secret instead of providing them as part of the ConfigMap

SrishT:

### Change log description
The Pravega Operator should enable ingestion of environment variables into the SegmentStore container by allowing users to create a custom config map and secret and associate the same with the segmentstore in the Pravega Spec.

### Purpose of the change
Fixes #332 

### What the code does
It allows the user to provide the name of the config map `segmentStoreEnvVars`  and the name of the secret `segmentStoreSecret` which contain all the values that need to be ingested into SegmentStore container. The secret  can be stored as an environment variable in the segment store container or it can be mounted to a volume depending on whether the user provides a value to the field `mountPath` or leaves it empty.

### How to verify it
The key-value pairs that have been added to the config map created by the user should be added to the list of environment variables of the segment store pods. The secret on the contrary can be added in two different ways. If the value `mountPath` within the `segmentStoreSecret` is empty, the secret will be added to the list of environment variables of the segment store pods, whereas if `mountPath` is non-empty, the secret will be mounted to a volume at the path that has been specified.

### DIFF
--- a/pkg/controller/pravega/pravega_segmentstore.go
+++ b/pkg/controller/pravega/pravega_segmentstore.go
@@ -66,8 +66,8 @@ func MakeSegmentStorePodTemplate(p *api.PravegaCluster) corev1.PodTemplateSpec {
 }
 
 func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
-	configMapName := util.GetConfigMapNameForSSEnvVars()
-	secretName := p.Spec.Pravega.SegmentStoreSecret.Secret
+	configMapName := strings.TrimSpace(p.Spec.Pravega.SegmentStoreEnvVars)
+	secret := p.Spec.Pravega.SegmentStoreSecret
 	environment := []corev1.EnvFromSource{
 		{
 			ConfigMapRef: &corev1.ConfigMapEnvSource{
@@ -77,7 +77,7 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 			},
 		},
 	}
-	if len(configMapName) > 0 {
+	if configMapName != "" {
 		environment = append(environment, corev1.EnvFromSource{
 			ConfigMapRef: &corev1.ConfigMapEnvSource{
 				LocalObjectReference: corev1.LocalObjectReference{
@@ -86,11 +86,11 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 			},
 		})
 	}
-	if len(secretName) > 0 && !p.Spec.Pravega.SegmentStoreSecret.MountToVolume {
+	if strings.TrimSpace(secret.Secret) != "" && strings.TrimSpace(secret.MountPath) == "" {
 		environment = append(environment, corev1.EnvFromSource{
 			SecretRef: &corev1.SecretEnvSource{
 				LocalObjectReference: corev1.LocalObjectReference{
-					Name: secretName,
+					Name: strings.TrimSpace(secret.Secret),
 				},
 			},
 		})
@@ -328,21 +328,21 @@ func configureTier2Filesystem(podSpec *corev1.PodSpec, pravegaSpec *api.PravegaS
 }
 
 func configureSegmentstoreSecret(podSpec *corev1.PodSpec, p *api.PravegaCluster) {
-	secretName := p.Spec.Pravega.SegmentStoreSecret.Secret
-	if len(secretName) > 0 && p.Spec.Pravega.SegmentStoreSecret.MountToVolume {
+	secret := p.Spec.Pravega.SegmentStoreSecret
+	if strings.TrimSpace(secret.Secret) != "" && strings.TrimSpace(secret.MountPath) != "" {
 		vol := corev1.Volume{
-			Name: p.Spec.Pravega.SegmentStoreSecret.VolumeName,
+			Name: ssSecretVolumeName,
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
-					SecretName: secretName,
+					SecretName: strings.TrimSpace(secret.Secret),
 				},
 			},
 		}
 		podSpec.Volumes = append(podSpec.Volumes, vol)
 
 		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, corev1.VolumeMount{
-			Name:      p.Spec.Pravega.SegmentStoreSecret.VolumeName,
-			MountPath: p.Spec.Pravega.SegmentStoreSecret.VolumeMountPath,
+			Name:      ssSecretVolumeName,
+			MountPath: strings.TrimSpace(secret.MountPath),
 		})
 	}
 }
--- a/pkg/controller/pravega/pravega_segmentstore.go
+++ b/pkg/controller/pravega/pravega_segmentstore.go
@@ -158,7 +158,7 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 	return podSpec
 }
 
-func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
+func MakeSegmentstoreConfigMap(p *api.PravegaCluster, cm *corev1.ConfigMap) *corev1.ConfigMap {
 	javaOpts := []string{
 		"-Dpravegaservice.clusterName=" + p.Name,
 	}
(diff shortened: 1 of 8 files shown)


pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -158,7 +158,7 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 	return podSpec
 }
 
-func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
+func MakeSegmentstoreConfigMap(p *api.PravegaCluster, cm *corev1.ConfigMap) *corev1.ConfigMap {
```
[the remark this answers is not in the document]
pbelgundi: If `SegmentStoreConfigMap` variable is non-nil, it can be directly added as one more element in the corev1.EnvFromSource Array:
https://github.com/pravega/pravega-operator/blob/3250498bd927c7a0df5af52394bcf94309f288d8/pkg/controller/pravega/pravega_segmentstore.go#L70

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -239,7 +239,13 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
 		return err
 	}
 
-	configMap := pravega.MakeSegmentstoreConfigMap(p)
+	cm := &corev1.ConfigMap{}
+	name := p.Spec.Pravega.SegmentStoreConfigMap
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, cm)
```
pbelgundi: This code is not needed. Please add directly to podspec as stated above. There is no need to confirm if the config map exists. If it does not deployment will fail and that is the expected behavior.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -342,7 +342,13 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
 		// This will trigger the rolling upgrade process
 		log.Printf("updating statefulset (%s) template image to '%s'", sts.Name, targetImage)
 
-		configMap := pravega.MakeSegmentstoreConfigMap(p)
+		cm := &corev1.ConfigMap{}
```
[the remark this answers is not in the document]
pbelgundi: There should be no impact to upgrades.

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -75,6 +77,24 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
 			},
 		},
 	}
+	if len(configMapName) > 0 {
```
pbelgundi: Why a length check, what if user enters a name that is 1 blank space? Could we do a string.trim() and then check for empty string "" ?

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -75,6 +77,24 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
...
+	}
+	if len(secretName) > 0 && !p.Spec.Pravega.SegmentStoreSecret.MountToVolume {
+		environment = append(environment, corev1.EnvFromSource{
+			SecretRef: &corev1.SecretEnvSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: secretName,
+				},
+			},
+		})
+	}
```
pbelgundi: Same comment as len(configMapName)

pbelgundi: Do you think this bool variable -`p.Spec.Pravega.SegmentStoreSecret.MountToVolume` is really necessary? If VolumeMountPath is not specified, but SecretName is specified, we could assume they want to expose as envvars... else use mountPath

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -305,6 +327,26 @@ func configureTier2Filesystem(podSpec *corev1.PodSpec, pravegaSpec *api.PravegaS
 	}
 }
 
+func configureSegmentstoreSecret(podSpec *corev1.PodSpec, p *api.PravegaCluster) {
+	secretName := p.Spec.Pravega.SegmentStoreSecret.Secret
+	if len(secretName) > 0 && p.Spec.Pravega.SegmentStoreSecret.MountToVolume {
```
pbelgundi: No length checks please.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -239,12 +239,20 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
...
 	configMap := pravega.MakeSegmentstoreConfigMap(p)
 	controllerutil.SetControllerReference(p, configMap, r.scheme)
 	err = r.client.Create(context.TODO(), configMap)
 	if err != nil && !errors.IsAlreadyExists(err) {
 		return err
 	}
+	flag := r.compareKeys(cm, configMap)
+	if !flag {
+		log.Println("Rejecting the environment variables provided")
```
pbelgundi: Please log as Error().

pbelgundi: Also state which Key caused the clash. Otherwise user does not know what went wrong.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -559,3 +567,15 @@ func (r *ReconcilePravegaCluster) isRollbackTriggered(p *pravegav1alpha1.Pravega
 	}
 	return false
 }
+
+func (r *ReconcilePravegaCluster) compareKeys(cm1 *corev1.ConfigMap, cm2 *corev1.ConfigMap) bool {
```
pbelgundi: This should be a method that returns intersection of set A and Set B. Return a `string` and `error` where the `string` returns all overlapping keys and error is non-nil if we failed to compute the intersection.

pkg/util/pravegacluster.go
```diff
@@ -81,6 +82,18 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
 	return fmt.Sprintf("%s-pravega-segmentstore", clusterName)
 }
 
+func SetConfigMapNameForSSEnvVars(name string, flag bool) {
```
pbelgundi: These setter/getter methods are not needed.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -239,12 +239,20 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
...
 	controllerutil.SetControllerReference(p, configMap, r.scheme)
 	err = r.client.Create(context.TODO(), configMap)
 	if err != nil && !errors.IsAlreadyExists(err) {
 		return err
 	}
+	flag := r.compareKeys(cm, configMap)
+	if !flag {
+		log.Println("Rejecting the environment variables provided")
+	}
+	util.SetConfigMapNameForSSEnvVars(name, flag)
```
pbelgundi: We should **not** be updating any global variables. You could create a new field under PravegaCluster Status to indicate if ConfigMap Validation passed or failed and use that to decide if the CM should be added to the pod or not. The correct way to do the validation is actually to use a validationWebHook, but that may take longer to implement, so for now we should be good with this...

pkg/util/pravegacluster.go
pbelgundi: Besides its incorrect to use "global" variables like ssEnvVar in operator code as it would interfere with managing multiple Pravega Clusters.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -239,12 +239,20 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
...
+	cm := &corev1.ConfigMap{}
+	name := p.Spec.Pravega.SegmentStoreEnvVars
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, cm)
 	configMap := pravega.MakeSegmentstoreConfigMap(p)
 	controllerutil.SetControllerReference(p, configMap, r.scheme)
 	err = r.client.Create(context.TODO(), configMap)
 	if err != nil && !errors.IsAlreadyExists(err) {
 		return err
 	}
+	flag := r.compareKeys(cm, configMap)
```
pbelgundi: Could we create a new method `validateUserConfigMap()` and do the get on user supplied config map and compareKeys inside the same?

PR #335, opened 2020-03-07
Issue 312 making ss cache volumes optional

### LINKED ISSUE #312: Make Segment Store cache volume optional for Pravega +0.7
### Description
In Pravega 0.7, the Segment Store cache is completely in-memory. This means that the PVC that the operator creates is not necessary anymore. We need to make the deployment of this volume optional, depending on whether the version of Pravega is older or newer than 0.7.

### Importance
Must have for 0.7.

### Location
Segment Store PVC creation.

We need to consider whether if we want to make this as a flag in the Operator manifest or whether the Operator should automatically look at the Pravega version and deploy the volume or not based on that.

Prabhaker24:
Signed-off-by: prabhaker24 Prabhaker.Saxena@dell.com

## Change log description
From Pravega 0.7, the Segment Store cache is completely in-memory. This means that the PVC that the operator creates is not necessary anymore. We need to make the deployment of this volume optional, depending on whether the version of Pravega is older or newer than 0.7, thus these changes will deploy as well as upgrade the version accordingly so as to not include pvc's incase we are deploying or upgrading to a Pravega version 0.7 or above and remove the old ones deployed by a version below 0.7 in case of an upgrade.
In Case of a rollback when an upgrade from a version above 0.7 to a version below 0.7 fails, it should remove the pods of the segment store above version 0.7 and should deploy all segment store pods of version below 0.7 along with their pvc's

## Purpose of the change
Fixes #312 

## What the code does
These changes will deploy the segment store with the pvc's for a version below 0.7 and will deploy the segment store without pvc's for a version 0.7 or above, it will also take care of upgrades in case the upgrades are from a version below 0.7 to a version above or equal to 0.7 and in this case it will remove the pvc's created by the older version while deploying the segment store for the version above 0.7.  
It will also handle the rollback in case of an upgrade failure from a version below 0.7 to version above 0.7 and will display appropriate upgrade failed message and when the user changes the version to the last version(below 0.7 from which he was upgrading) it will rollback it to this below 0.7 version and remove the old pods of the segment store having version above 0.7.
It will also display appropriate user message incase the above rollback also fails.

## How to verify it
1. Deploy Pravega with a version 0.7 or above and it should be deployed without any pvc's
2. Upgrade Pravega with a version blow 0.7 to a version above 0.7 it should do two this deploy the new segment store without pvc's and remove the old segment store as well as the pvc's attached to it.
3. Upgrade Pravega from a version below 0.7 to a version above 0.7 
4. Upgrade Pravega from a version above 0.7 to a version above 0.7
5. Upgrade from 0.6.0 to 0.6.2 . 
6. I/o should be working in case of all the above four processes both while the upgrade or deployment is in the process as well as when it's completed to check that Benchmark can be run.
7. Rollback in case the upgrade fails while upgrading from a version below 0.7 to a  version below 0.7
8.  Rollback in case the upgrade fails while upgrading from a version below 0.7 to a  version above 0.7
9. Rollback in case the upgrade fails while upgrading from a version above 0.7 to a  version above 0.7
10. Displays appropriate message in case rollback fails while doing a rollback if an upgrade fails  from a version below 0.7 to a  version above 0.7 .


pkg/controller/pravegacluster/upgrade.go
```diff
@@ -322,10 +322,78 @@ func (r *ReconcilePravegaCluster) syncControllerVersion(p *pravegav1alpha1.Prave
 	return true, nil
 }
 
+func (r *ReconcilePravegaCluster) syncStoreVersion(p *pravegav1alpha1.PravegaCluster) (synced bool, err error) {
+	//this is to check if the target image is above 07 and the current image is below 07
+	if p.Status.TargetVersion != "" && !util.IsVersionBelow07(p.Status.TargetVersion) && util.IsVersionBelow07(p.Status.CurrentVersion) {
```
pbelgundi: Replace these 3 checks with p.Status.IsClusterUpgradingTo07()

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -322,10 +322,78 @@ func (r *ReconcilePravegaCluster) syncControllerVersion(p *pravegav1alpha1.Prave
...
+		return r.syncSegmentStoreVersionFrom06To07(p)
+	}
+	return r.syncSegmentStoreVersion(p)
+}
+
+func (r *ReconcilePravegaCluster) syncSegmentStoreVersionFrom06To07(p *pravegav1alpha1.PravegaCluster) (synced bool, err error) {
```
pbelgundi: Could we call this syncSegmentStoreVersionTo07()

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -322,10 +322,78 @@ func (r *ReconcilePravegaCluster) syncControllerVersion(p *pravegav1alpha1.Prave
...
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: newsts.Name, Namespace: p.Namespace}, newsts)
+		//this checks is to verify if the old sts is not present new sts should be present thus this ends upgrade and true is returned
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+
+	//this check is run till the value of old sts replicas is greater than 0 and will increase two replicas of the new sts and delete 2 replicas of the old sts
+	if *oldsts.Spec.Replicas > 0 {
```
pbelgundi: Can we not check for <2 instead of 0....
If # replicas in oldsts is <=2, then just increase the replicas in NewSTS by 2 and delete the old STS. Before deleting old STS invoke `r.syncStatefulSetPvc(oldsts)`

pbelgundi:
@Prabhaker24  in case of scenarios tested by you, please specify what was the expected outcome and weather you were able to see the expected outcome in your tests.
Here is the list of scenarios you need to cover:
1. Deploy Pravega with version < 0.7 and it should be deployed successfully with Volume templates attached to SSS pods.
2. Deploy Pravega with version >=0.7 and and it should be deployed successfully without PVCs attached to SSS STS.
3. Upgrade from 0.6 to 0.7
4. Upgrade from 0.7.1 to 0.7.2
5. Upgrade from 0.6.0 to 0.6.2.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -301,9 +308,13 @@ func (r *ReconcilePravegaCluster) syncClusterSize(p *pravegav1alpha1.PravegaClus
 		return err
 	}
 
-	err = r.syncSegmentStoreSize(p)
-	if err != nil {
-		return err
+	/*this condition is to stop syncSegmentstore version from running when we are updating the segment store version as in case of
+	updapting the ss from version below 06 to version above 07 get() call in syncSegmentstore will result in error */
+	if p.Status.CurrentVersion == p.Spec.Version {
```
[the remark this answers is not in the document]
Prabhaker24: This is needed to avoid client.get() inside the syncSegmentstore version() as that will result in error while upgrading to a version above 07 as it will get the lastest name of the ss which is not yet created as the creation code is inside the upgrade.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -170,12 +170,16 @@ func (r *ReconcilePravegaCluster) deployCluster(p *pravegav1alpha1.PravegaCluste
 		return err
 	}
 
-	err = r.deploySegmentStore(p)
-	if err != nil {
-		log.Printf("failed to deploy segment store: %v", err)
-		return err
+	_, upgradeCondition := p.Status.GetClusterCondition(pravegav1alpha1.ClusterConditionUpgrading)
+	/*this check is to avoid creation of a new segmentstore when the CurrentVersionis 06 and target version is 07
```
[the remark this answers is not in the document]
Prabhaker24: I have added a function isClusterUpgradingTo07() and used it.

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: I have changed this.

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -170,12 +170,16 @@ func (r *ReconcilePravegaCluster) deployCluster(p *pravegav1alpha1.PravegaCluste
...
+	  as we are doing it in the upgrade path*/
+	if upgradeCondition == nil {
+		err = r.deploySegmentStore(p)
+		if err != nil {
+			log.Printf("failed to deploy segment store: %v", err)
```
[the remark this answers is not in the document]
Prabhaker24: I haven't added this error

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -158,6 +154,27 @@ func makeSegmentstorePodSpec(p *api.PravegaCluster) corev1.PodSpec {
...
+				Name:      cacheVolumeName,
+				MountPath: cacheVolumeMountPoint,
+			},
+			{
+				Name:      heapDumpName,
+				MountPath: heapDumpDir,
+			},
+		}
+	}
+	return []corev1.VolumeMount{
```
pbelgundi: Lets write more neat and readable code. First create a VolumeMount slice with heapdump mount in it. If util.IsVersionBelow07(p.Spec.Version) append cacheVolumeMount to the slice

pkg/util/pravegacluster.go
```diff
@@ -81,8 +82,32 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
+func IsVersionBelow07(ver string) bool {
+	first3 := strings.Trim(ver, "\t \n")[0:3]
+	if first3 == "0.6" || first3 == "0.5" || first3 == "0.4" || first3 == "0.3" || first3 == "0.2" || first3 == "0.1" {
+		return true
+	}
+	return false
+}
+
+func IsClusterUpgradingTo07(p *api.PravegaCluster) bool {
+	/*this check is to avoid creation of a new segmentstore when the CurrentVersionis 06 and target version is 07
```
pbelgundi: Please provide appropriate comments. "This check is to determine if we are upgrading from a version below 0.7 to version 0.7 or later." Also comments should be provided above the function definition and not inside. Lets use standard coding conventions please.

pkg/util/pravegacluster.go
```diff
@@ -81,8 +82,32 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
+	  as we are doing it in the upgrade path*/
+	if !IsVersionBelow07(p.Spec.Version) && IsVersionBelow07(p.Status.CurrentVersion) {
+		return true
+	}
+	return false
+}
+
+//to return name of segmentstore based on the version
+func StatefulSetNameForSegmentstore(p *api.PravegaCluster) string {
+	//if version is below 0.7 this name will be assigned
```
pbelgundi: Please provide meaningful comments.
For example:
"StatefulSet name when cluster version >=0.7"

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -322,10 +322,82 @@ func (r *ReconcilePravegaCluster) syncControllerVersion(p *pravegav1alpha1.Prave
...
+	}
+	return r.syncSegmentStoreVersion(p)
+}
+
+func (r *ReconcilePravegaCluster) syncSegmentStoreVersionTo07(p *pravegav1alpha1.PravegaCluster) (synced bool, err error) {
+	newsts := pravega.MakeSegmentStoreStatefulSet(p)
+	controllerutil.SetControllerReference(p, newsts, r.scheme)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: newsts.Name, Namespace: p.Namespace}, newsts)
+	//this check is to see if the newsts is present or not if it's not present it will be created here
+	if err != nil {
```
pbelgundi: to check if the STS exists or not, you should be checking using:
``` if err != nil  && errors.IsNotFound(err) ```

pkg/util/pravegacluster.go
Prabhaker24: changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/util/pravegacluster.go
Prabhaker24: done

pbelgundi:
Changes migrated to PR: #342

PR #341, opened 2020-03-16
Issue 303: Providing the supportedVersions map as a ConfigMap

### LINKED ISSUE #303: Remove the supported versions map used by the webhook to validate an upgrade
### Description
Before installation of a Pravega Cluster resource, the webhook validates whether the version is supported or not by doing a lookup in the `supportedVersions` map maintained. Even an upgrade from a version x to a version y is allowed only if y is a value against the key x in this map. The problem with this approach is that we need to keep adding entries with every new release which is a very inefficient way of handling this.

### Location
`webhook.go` file under `pkg` folder.

SrishT:

### Change log description
The supportedVersions map, which is used to validate Pravega Cluster installations and upgrades from one version to another is currently hardcoded in the webhook code. This change moves this hardcoded map out of the webhook code.

### Purpose of the change
Fixes #303 

### What the code does
It provides the version map as a configmap instead of it being hardcoded in the webhook code.

### How to verify it
The request for installation of a Pravega Cluster resource should be blocked by the webhook if the version field does not contain a valid value. Also, upgrade from a version x to y should be allowed if y lies in the upgrade path of x, and should be disallowed otherwise.


charts/pravega-operator/values.yaml
```diff
@@ -22,3 +22,6 @@ crd:
 
 # whether to enable test mode
 testmode: false
+
+# whether to run in production environment
+production: true
```
pbelgundi: What is this for?

charts/pravega/values.yaml
```diff
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-version: 0.5.0
+version: latest
```
pbelgundi: "latest" version is not accepted by webhook in it validation. version can be changed to `0.7.0`

pkg/util/pravegacluster.go
```diff
@@ -33,6 +33,10 @@ func init() {
 	versionRegexp = regexp.MustCompile(MajorMinorVersionRegexp)
 }
 
+func ConfigMapNameForOperatorVersions() string {
```
pbelgundi: This is not a config map for "operator versions" It is a map for "pravega versions" Please provide a suitable name....

pkg/webhook/webhook.go
```diff
@@ -117,7 +108,7 @@ func (pwh *pravegaWebhookHandler) mutatePravegaVersion(ctx context.Context, p *p
 	if err != nil {
 		return fmt.Errorf("request version is not in valid format: %v", err)
 	}
-	if _, ok := supportedVersions[normRequestVersion]; !ok {
+	if _, ok := versionMap[normRequestVersion]; !ok {
```
pbelgundi: It would be good to maintain the same reference name "supportedVersions" for the Map, so do not need to change the references everywhere.

charts/pravega-operator/values.yaml
SrishT: This value is to determine which of the 2 config-maps to deploy : the one for production env or the one for test env.

pbelgundi: That can be done using `testmode` variable.

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
```diff
@@ -25,6 +25,10 @@ const (
 	// DefaultServiceType is the default service type for external access
 	DefaultServiceType = v1.ServiceTypeLoadBalancer
 
+	// DefaultVersionMap is the default configMap which contains the list of
+	// supported versions for the pravega cluster
+	DefaultVersionMap = "version-map"
```
pbelgundi: Could we call this "upgrade_supported_versions" ?

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
```diff
@@ -103,6 +107,10 @@ type ClusterSpec struct {
 	//pravega-bookie-2.pravega-bookie-headless.default:3181
 	BookkeeperUri string `json:"bookkeeperUri"`
 
+	// VersionMap is the name of the configMap which contains the list of
+	// supported versions for the pravega cluster
+	VersionMap string `json:"versionMap"`
```
pbelgundi: This is really a upgrade_supported_versions map. Variable name should reflect that.

SrishT:
> Please make sure to test these changes with "testmode:false" and "testmode:true" in the chart values.yaml file.

Both these modes have been tested
`testmode:true` deploys the ConfigMap for test environment while `testmode:false` deploys the ConfigMap for the production environment

charts/pravega/values.yaml
```diff
@@ -24,3 +24,5 @@ pravega:
   storage:
     className: standard
   tier2: pravega-tier2
+
+testmode: true
```
pbelgundi: We don't need to add a new variable to the pravega chart here. The pravega-operator has a "testmode" variable, if that is set to true, we should use the test config map yaml else use the prod one.
https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator/values.yaml#L24

pkg/apis/pravega/v1alpha1/pravegacluster_types.go
```diff
@@ -103,6 +107,10 @@ type ClusterSpec struct {
 	//pravega-bookie-2.pravega-bookie-headless.default:3181
 	BookkeeperUri string `json:"bookkeeperUri"`
 
+	// SupportedUpgradePaths is the name of the configMap which contains the list of
+	// supported upgrade paths for all versions of the pravega cluster
+	SupportedUpgradePaths string `json:"supportedUpgradePaths"`
```
pbelgundi: I don't see the need for this variable or its default value. We can have a **const** defined for the config-map name and use that. In any Pravega installation user would be required to create only **1 config map** for checking supported upgrade paths and we can mandate that this config map must be created with this name. Please choose a suitable name like "pravega-supported-upgrade-paths".  We need to explain this in the upgrade documentation as well.

pkg/webhook/webhook.go
```diff
@@ -88,6 +70,15 @@ func (pwh *pravegaWebhookHandler) mutatePravegaManifest(ctx context.Context, p *
 }
 
 func (pwh *pravegaWebhookHandler) mutatePravegaVersion(ctx context.Context, p *pravegav1alpha1.PravegaCluster) error {
+	configMap := &corev1.ConfigMap{}
+	err := pwh.client.Get(ctx, types.NamespacedName{Name: p.Spec.SupportedUpgradePaths, Namespace: p.Namespace}, configMap)
+	if err != nil {
```
pbelgundi: It is important to distinguish between error in client.Get and exists check for config map.
Here is how we can check if the config map exists:
https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravegacluster/pravegacluster_controller.go#L93

If config map is not found, please provide a suitable message to the user to create this config map. Current message does not explain if the user can do anything to fix this problem.

pkg/webhook/webhook.go
```diff
@@ -88,6 +70,15 @@ func (pwh *pravegaWebhookHandler) mutatePravegaManifest(ctx context.Context, p *
...
+		log.Println(err)
```
pbelgundi: Why are we logging as Println. We should log this with loglevel error.

SrishT: done

pkg/webhook/webhook.go
SrishT: updated

test/e2e/resources/version-configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+  0.3.2: "0.3.2"
+  0.4.0: "0.4.0"
+  0.5.0: "0.5.0,0.5.1,0.6.0,0.6.1,0.6.2,0.6.3,0.7.0,0.8.0"
+  0.5.1: "0.5.1,0.6.0,0.6.1,0.6.2,0.6.3,0.7.0,0.8.0"
+  0.6.0: "0.6.0,0.6.1,0.6.2,0.6.3,0.7.0,0.8.0"
+  0.6.1: "0.6.1,0.6.2,0.6.3,0.7.0,0.8.0"
+  0.6.2: "0.6.2,0.6.3,0.7.0,0.8.0"
+  0.6.3: "0.6.3,0.7.0,0.8.0"
+  0.7.0: "0.7.0,0.8.0"
+  0.8.0: "0.8.0"
```
pbelgundi: Does this map need to have 0.8.0?? I guess not.

pbelgundi: We are using the same hardcoded map values in 3-4 different places. Tomorrow when a new release is created we would need to update in all these 3-4 places. Can we do better?

deploy/version_map.yaml
```diff
@@ -0,0 +1,17 @@
...
+data:
+  0.1.0: "0.1.0"
+  0.2.0: "0.2.0"
+  0.3.0: "0.3.0,0.3.1,0.3.2"
+  0.3.1: "0.3.1,0.3.2"
+  0.3.2: "0.3.2"
+  0.4.0: "0.4.0"
+  0.5.0: "0.5.0,0.5.1,0.6.0,0.6.1,0.6.2,0.7.0"
+  0.5.1: "0.5.1,0.6.0,0.6.1,0.6.2,0.7.0"
+  0.6.0: "0.6.0,0.6.1,0.6.2,0,0.7.0"
```
pbelgundi: a typo here:
0.6.0: "0.6.0,0.6.1,0.6.2,**0**,0.7.0"

charts/pravega/templates/version_map.yaml
```diff
@@ -0,0 +1,17 @@
...
+data:
+  0.1.0: "0.1.0"
+  0.2.0: "0.2.0"
+  0.3.0: "0.3.0,0.3.1,0.3.2"
+  0.3.1: "0.3.1,0.3.2"
+  0.3.2: "0.3.2"
+  0.4.0: "0.4.0"
+  0.5.0: "0.5.0,0.5.1,0.6.0,0.6.1,0.6.2,0.7.0"
+  0.5.1: "0.5.1,0.6.0,0.6.1,0.6.2,0.7.0"
+  0.6.0: "0.6.0,0.6.1,0.6.2,0,0.7.0"
```
pbelgundi: typo in line "0.6.0: "0.6.0,0.6.1,0.6.2,**0**,0.7.0""

deploy/version_map.yaml
SrishT: fixed

charts/pravega/templates/version_map.yaml
SrishT: fixed

test/e2e/resources/version-configmap.yaml
SrishT: The map is now hardcoded in 2 places based on whether the user wants to install the cluster manually or using charts. The map in the test file does not need to be updated.

PR #342, opened 2020-03-16
Issue 312 making ss cache volumes optional

### LINKED ISSUE #312: Make Segment Store cache volume optional for Pravega +0.7

Prabhaker24:
Signed-off-by: prabhaker24 Prabhaker.Saxena@dell.com

## Change log description
From Pravega 0.7, the Segment Store cache is completely in-memory. This means that the PVC that the operator creates is not necessary anymore. We need to make the deployment of this volume optional, depending on whether the version of Pravega is older or newer than 0.7, thus these changes will deploy as well as upgrade the version accordingly so as to not include pvc's incase we are deploying or upgrading to a Pravega version 0.7 or above and remove the old ones deployed by a version below 0.7 in case of an upgrade.
In Case of a rollback when an upgrade from a version above 0.7 to a version below 0.7 fails, it should remove the pods of the segment store above version 0.7 and should deploy all segment store pods of version below 0.7 along with their pvc's

## Purpose of the change
Fixes #312 

## What the code does
These changes will deploy the segment store with the pvc's for a version below 0.7 and will deploy the segment store without pvc's for a version 0.7 or above, it will also take care of upgrades in case the upgrades are from a version below 0.7 to a version above or equal to 0.7 and in this case it will remove the pvc's created by the older version while deploying the segment store for the version above 0.7.  
It will also handle the rollback in case of an upgrade failure from a version below 0.7 to version above 0.7 and will display appropriate upgrade failed message and when the user changes the version to the last version(below 0.7 from which he was upgrading) it will rollback it to this below 0.7 version and remove the old pods of the segment store having version above 0.7.
It will also display appropriate user message incase the above rollback also fails.

## How to verify it
1. Deploy Pravega with a version 0.7 or above and it should be deployed without any pvc's
2. Upgrade Pravega with a version blow 0.7 to a version above 0.7 it should do two this deploy the new segment store without pvc's and remove the old segment store as well as the pvc's attached to it.
3. Upgrade Pravega from a version below 0.7 to a version above 0.7 
4. Upgrade Pravega from a version above 0.7 to a version above 0.7
5. Upgrade from 0.6.0 to 0.6.2 . 
6. I/o should be working in case of all the above four processes both while the upgrade or deployment is in the process as well as when it's completed to check that Benchmark can be run.
7. Rollback in case the upgrade fails while upgrading from a version below 0.7 to a  version below 0.7
8.  Rollback in case the upgrade fails while upgrading from a version below 0.7 to a  version above 0.7
9. Rollback in case the upgrade fails while upgrading from a version above 0.7 to a  version above 0.7
10. Displays appropriate message in case rollback fails while doing a rollback if an upgrade fails from a version below 0.7 to a  version above 0.7.


pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -50,9 +51,13 @@ func MakeSegmentStoreStatefulSet(pravegaCluster *api.PravegaCluster) *appsv1.Sta
 			Selector: &metav1.LabelSelector{
 				MatchLabels: util.LabelsForSegmentStore(pravegaCluster),
 			},
-			VolumeClaimTemplates: makeCacheVolumeClaimTemplate(pravegaCluster.Spec.Pravega),
 		},
 	}
+	if util.IsVersionBelow07(pravegaCluster.Spec.Version) {
+		log.Printf("creating volume claim ")
```
pbelgundi: Either print a more meaningful message or remove this log

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -159,10 +159,14 @@ func (r *ReconcilePravegaCluster) deployCluster(p *pravegav1alpha1.PravegaCluste
 		return err
 	}
 
-	err = r.deploySegmentStore(p)
-	if err != nil {
-		log.Printf("failed to deploy segment store: %v", err)
-		return err
+	/*this check is to avoid creation of a new segmentstore when the CurrentVersionis is below 07 and target version is above 07
```
pbelgundi: Fix typo "CurrentVersionis"

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -249,9 +253,13 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
 }
 
 func (r *ReconcilePravegaCluster) syncClusterSize(p *pravegav1alpha1.PravegaCluster) (err error) {
-	err = r.syncSegmentStoreSize(p)
-	if err != nil {
-		return err
+	/*this condition is to stop syncSegmentstore version from running when we are updapting the ss from version below 07
```
pbelgundi: Could we just say in the comment " We skip calling syncSegmentStoreSize() during upgrade/rollback from version 07"

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -281,9 +289,12 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreSize(p *pravegav1alpha1.Praveg
 			return fmt.Errorf("failed to update size of stateful-set (%s): %v", sts.Name, err)
 		}
 
-		err = r.syncStatefulSetPvc(sts)
-		if err != nil {
-			return fmt.Errorf("failed to sync pvcs of stateful-set (%s): %v", sts.Name, err)
+		// this check is to avoid calling syncStatefulSetPvc() over ss with version above 07
```
pbelgundi: Same as above

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+//this function is to check are we doing a rollback in case of a upgrade failure while upgrading from a version below 07 to a version above 07
+func (r *ReconcilePravegaCluster) IsClusterRollbackingFrom07(p *pravegav1alpha1.PravegaCluster) bool {
+	if !r.isRollbackTriggered(p) {
+		return false
+	}
+	//checking if sts above07 exsists
+	stsAbove07 := &appsv1.StatefulSet{}
+	name := util.StatefulSetNameForSegmentstoreAbove07(p.Name)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, stsAbove07)
+	if err != nil && errors.IsNotFound(err) {
```
pbelgundi: What if client.Get() returns error, but it is not a "NotFound" error? Please add a check similar to :
https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravegacluster/pravegacluster_controller.go#L92

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+		return false
+	}
+	//checking if sts below07 exsists
+	stsBelow07 := &appsv1.StatefulSet{}
+	name = util.StatefulSetNameForSegmentstoreBelow07(p.Name)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, stsBelow07)
+	if err != nil && errors.IsNotFound(err) {
```
pbelgundi: Same as above.

pkg/controller/pravegacluster/upgrade.go
pbelgundi: All errors  must be logged with log level "**Error**". We're not logging anything here.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+		return false
+	}
+	//true will only be return if both sts are present
+	return true
+}
+
+//this function calls the syncstoreversion for upgrade and rollback operation
+func (r *ReconcilePravegaCluster) syncStoreVersion(p *pravegav1alpha1.PravegaCluster) (synced bool, err error) {
+	//in case of a upgrade from a version below 07 to a version above 07 syncSegmentStoreVersionTo07() is called and the util.IsClusterUpgradingTo07(p)  is true in this case
```
pbelgundi: Please remove these 2 comments they are unnecessary, code is self explanatory

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	//in case of a rollback when upgrade fails from a version below 07 to a version above 07 syncSegmentStoreVersionTo07() is called and the r.IsClusterRollbackingFrom07(p) is true in this case
+	if util.IsClusterUpgradingTo07(p) || r.IsClusterRollbackingFrom07(p) {
+		return r.syncSegmentStoreVersionTo07(p)
+	}
+	//for all other cases of upgrades and rollback this function is called
```
pbelgundi: remove comment

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	return r.syncSegmentStoreVersion(p)
+}
+
+//this fuction is called when we want to upgrade ss from a version below 07 to a version above 07  as well as when the such upgrades fail then for doing rollback also it's called
```
pbelgundi: This much is enough: "To handle upgrade/rollback from Pravega version < 0.7 to Pravega Version >= 0.7"

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+func (r *ReconcilePravegaCluster) syncSegmentStoreVersionTo07(p *pravegav1alpha1.PravegaCluster) (synced bool, err error) {
+	newsts := pravega.MakeSegmentStoreStatefulSet(p)
+	controllerutil.SetControllerReference(p, newsts, r.scheme)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: newsts.Name, Namespace: p.Namespace}, newsts)
+	//this check is to see if the newsts is present or not if it's not present it will be created here
+	if err != nil && errors.IsNotFound(err) {
```
pbelgundi: Same as above. Please log all errors. We should not be hiding them.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	}
+
+	p.Status.UpdateProgress(pravegav1alpha1.UpdatingSegmentstoreReason, "0")
+	//this check to ensure that the oldsts always decrease by 2 as well as newsts pods increase by 2 only then the next increment or decrement happen
+	//first condition is true and used in the case of rollbackCondition
+	//second condition is true and used in case of Upgrade
+	if (r.IsClusterRollbackingFrom07(p) && newsts.Status.ReadyReplicas == *newsts.Spec.Replicas) || (oldsts.Status.ReadyReplicas+newsts.Status.ReadyReplicas == p.Spec.Pravega.SegmentStoreReplicas && newsts.Status.ReadyReplicas == *newsts.Spec.Replicas) {
+		//this check is run till the value of old sts replicas is greater than 0 and will increase two replicas of the new sts and delete 2 replicas of the old sts
+		if *oldsts.Spec.Replicas > 2 {
+			*newsts.Spec.Replicas = *newsts.Spec.Replicas + 2
```
pbelgundi: Encapsulate the changes to newSTS in a function and invoke that here.
Ditto for old STS, so its easier to follow what is happening.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,139 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+				return false, fmt.Errorf("updating statefulset (%s) failed due to %v", newsts.Name, err)
+			}
+			*oldsts.Spec.Replicas = *oldsts.Spec.Replicas - 2
+			err = r.client.Update(context.TODO(), oldsts)
+			if err != nil {
+				return false, fmt.Errorf("updating statefulset (%s) failed due to %v", oldsts.Name, err)
+			}
+		} else {
+			//here we remove the pvc's attached with the old sts and deleted it when old sts replicas have become 0
+			*newsts.Spec.Replicas = p.Spec.Pravega.SegmentStoreReplicas
```
pbelgundi: Again encapsulate changes to old STS in a method and invoke the same....
One more thing, we need to increment the newSTS first and then do stuff for deleting the old sts, so that we err on side of caution.... and can have more redundancy than less....

pkg/util/pravegacluster.go
```diff
@@ -81,8 +82,42 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
 }
 
-func StatefulSetNameForSegmentstore(clusterName string) string {
-	return fmt.Sprintf("%s-pravega-segmentstore", clusterName)
+//function to check if the version is below 0.7 or not
+func IsVersionBelow07(ver string) bool {
+	if ver == "" {
+		return true
+	}
+	first3 := strings.Trim(ver, "\t \n")[0:3]
```
pbelgundi: Why can't we use CompareVersions here? Lets try to reuse existing methods wherever possible, Its is not good to use only way of doing things here and another way in the webhook.

pkg/util/pravegacluster.go
```diff
@@ -81,8 +82,42 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
+	if first3 == "0.6" || first3 == "0.5" || first3 == "0.4" || first3 == "0.3" || first3 == "0.2" || first3 == "0.1" {
+		return true
+	}
+	return false
+}
+
+//this function will return true only in case of upgrading from a version below 0.7 to a version above 0.7
```
pbelgundi: Fix the comment, "to pravega version 0.7 or later"

pkg/util/pravegacluster.go
```diff
@@ -81,8 +82,42 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
+func IsClusterUpgradingTo07(p *api.PravegaCluster) bool {
+	if !IsVersionBelow07(p.Spec.Version) && IsVersionBelow07(p.Status.CurrentVersion) {
+		return true
+	}
+	return false
+}
+
+//if version is above or equals to 0.7 this name will be assigned
+func StatefulSetNameForSegmentstoreAbove07(name string) string {
+	return fmt.Sprintf("%s-pravega-above-version-07-segmentstore", name)
```
pbelgundi: It would be good to make minimum changes to SS name.
Could we just call this "pravega-segment-store"???

pkg/controller/pravega/pravega_segmentstore.go
Prabhaker24: removed

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: done

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: done

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: done

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: i have changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/util/pravegacluster.go
Prabhaker24: changed it

pkg/util/pravegacluster.go
Prabhaker24: fixed

pkg/util/pravegacluster.go
Prabhaker24: changed

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,178 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
 	return false, nil
 }
 
+//this function is to check are we doing a rollback in case of a upgrade failure while upgrading from a version below 07 to a version above 07
+func (r *ReconcilePravegaCluster) IsClusterRollbackingFrom07(p *pravegav1alpha1.PravegaCluster) bool {
+	if !r.isRollbackTriggered(p) {
+		return false
+	}
+	//checking if sts above07 exsists
+	stsAbove07 := &appsv1.StatefulSet{}
```
pbelgundi: Why are we checking for existence of old/new STS to understand if rollback is triggered?
This does not look right...
I would expect this to be as simple as :
```
IsClusterRollbackingFrom07 (. . .) (. . .) {
if !r.isRollbackTriggered(p) {
		return false
	}
return true
}
```

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,178 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	//this check is to see if the old sts is present or not
+	if err != nil && errors.IsNotFound(err) {
+		//this is the condition where we are checking if the new ready ss pods are equal to the ss replicas thus meaning upgrade is completed
+		if newsts.Status.ReadyReplicas == p.Spec.Pravega.SegmentStoreReplicas {
+			return true, nil
+		}
+	}
+
+	//TO detect upgade/rolback faiure
+	if oldsts.Status.ReadyReplicas+newsts.Status.ReadyReplicas != p.Spec.Pravega.SegmentStoreReplicas {
```
pbelgundi: != check should be changed to < as discussed.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,178 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false
+		}
+		log.Printf("failed to get PravegaCluster: %v", err)
+		return false
+	}
+
+	//checking if sts below07 exsists
+	stsBelow07 := &appsv1.StatefulSet{}
```
pbelgundi: How are we dealing with situations where one STS is deleted and other is present?

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,178 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+//this function will check if furter increment or decrement in pods needed in case of upgrade to version0.7 or above
+func (r *ReconcilePravegaCluster) upgradeConditionFor07(p *pravegav1alpha1.PravegaCluster, newsts *appsv1.StatefulSet, oldsts *appsv1.StatefulSet) bool {
+	if oldsts.Status.ReadyReplicas+newsts.Status.ReadyReplicas == p.Spec.Pravega.SegmentStoreReplicas && newsts.Status.ReadyReplicas == *newsts.Spec.Replicas {
+		return true
+	}
+	return false
+}
+
+//this function will increase two replicas of the new sts and delete 2 replicas of the old sts everytime it's called
+func (r *ReconcilePravegaCluster) incrementWhenReplicasMoreThan2(p *pravegav1alpha1.PravegaCluster, newsts *appsv1.StatefulSet, oldsts *appsv1.StatefulSet) error {
```
pbelgundi: Could we call this method `scaleSegmentStoreSTS()`

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,178 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	*oldsts.Spec.Replicas = *oldsts.Spec.Replicas - 2
+	err = r.client.Update(context.TODO(), oldsts)
+	if err != nil {
+		return fmt.Errorf("updating statefulset (%s) failed due to %v", oldsts.Name, err)
+	}
+	return nil
+}
+
+//This function will remove the pvc's attached with the old sts and deleted it when old sts replicas have become 0
+func (r *ReconcilePravegaCluster) incrementWhenReplicasLessThan2(p *pravegav1alpha1.PravegaCluster, newsts *appsv1.StatefulSet, oldsts *appsv1.StatefulSet) error {
```
pbelgundi: Please use a better method name that is indicative of what the method does..... Perhaps transitionToNewSTS() ???

pkg/util/pravegacluster.go
```diff
@@ -81,8 +82,43 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
 
-func StatefulSetNameForSegmentstore(clusterName string) string {
-	return fmt.Sprintf("%s-pravega-segmentstore", clusterName)
+//function to check if the version is below 0.7 or not
+func IsVersionBelow07(ver string) bool {
+	if ver == "" {
+		return true
+	}
+	normVer, _ := NormalizeVersion(ver)
+	trimNormVer := normVer[0:3]
```
pbelgundi: Lets use compareVersions instead of picking fixed number of elements from the normVer Array. That would be more elegant.

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: i have changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: that's not possible as we are not deleting the oldsts  till the time new sts hasn't come up completely meaning all it's ready replicas are equal to segmentstore replicas

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: we need to check the existence of both sts as we need to call the method syncSegmentStoreVersionTo07()  in case of rollback only when an upgrade from a version below 07 to a version above 07 has failed meaning both sts are there and as well as rollback is triggered so i am checking the presence of both sts.

pkg/util/pravegacluster.go
Prabhaker24: I need to pick the first 3 characters only as otherwise if I am checking for version after normalization it can be 0.6.0 0r 0.6.1 or 0.6.2 thus it will be a very long condition now I can just check if it's equal to 0.6 and it will cover all the 0.6.x type of cases

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: Changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: Changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: done

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -281,9 +288,12 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreSize(p *pravegav1alpha1.Praveg
 			return fmt.Errorf("failed to update size of stateful-set (%s): %v", sts.Name, err)
 		}
 
-		err = r.syncStatefulSetPvc(sts)
-		if err != nil {
-			return fmt.Errorf("failed to sync pvcs of stateful-set (%s): %v", sts.Name, err)
+		/*We skip calling syncStatefulSetPvc() during upgrade/rollback from version 07*/
+		if !util.IsClusterUpgradingTo07(p) && !r.isRollbackTriggered(p) {
```
pbelgundi: Same as above. Why can't we use "IsClusterRollbackingFrom07()"?

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -249,9 +253,12 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1alpha1.PravegaC
 }
 
 func (r *ReconcilePravegaCluster) syncClusterSize(p *pravegav1alpha1.PravegaCluster) (err error) {
-	err = r.syncSegmentStoreSize(p)
-	if err != nil {
-		return err
+	/*We skip calling syncSegmentStoreSize() during upgrade/rollback from version 07*/
+	if !util.IsClusterUpgradingTo07(p) && !r.isRollbackTriggered(p) {
```
pbelgundi: We should be checking if rollback from 07 is triggered. Here we're checking if any rollback is triggered. Why can't we use "IsClusterRollbackingFrom07()"?

pkg/controller/pravegacluster/upgrade.go
pbelgundi: The objective of this method is to figure out if the cluster is rolling back from 07 to a version below 07 and it should do just that and nothing else.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,180 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+		if errors.IsNotFound(err) {
+			return false
+		}
+		log.Printf("failed to get PravegaCluster: %v", err)
+		return false
+	}
+	return true
+}
+
+//To handle upgrade/rollback from Pravega version < 0.7 to Pravega Version >= 0.7
```
pbelgundi: Wrong comment on method

pkg/controller/pravegacluster/upgrade.go
pbelgundi: comment not removed yet.

pkg/controller/pravegacluster/upgrade.go
pbelgundi: I don;t see the error being logged????

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: I have changed it.

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: I have changed it

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: I have changed it

pkg/controller/pravegacluster/upgrade.go
Prabhaker24: removed

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -404,6 +405,174 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.Pra
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			*newsts.Spec.Replicas = 0
+			err2 := r.client.Create(context.TODO(), newsts)
+			if err2 != nil {
+				log.Printf("failed to create StatefulSet: %v", err)
+				return false, err
+			}
+		}
+		if !errors.IsNotFound(err) {
```
pbelgundi: This if is not needed, please remove.

Prabhaker24: Done

pkg/util/pravegacluster.go
```diff
@@ -85,8 +86,42 @@ func ConfigMapNameForSegmentstore(clusterName string) string {
...
+func IsClusterUpgradingTo07(p *api.PravegaCluster) bool {
+	if !IsVersionBelow07(p.Spec.Version) && IsVersionBelow07(p.Status.CurrentVersion) {
+		return true
+	}
+	return false
+}
+
+//if version is above or equals to 0.7 this name will be assigned
+func StatefulSetNameForSegmentstoreAbove07(name string) string {
+	return fmt.Sprintf("%s-pravega-segment-store", name)
```
jkhalack: This change just broke the upgrade.

jkhalack: `0.7.0` pravega clusters historically were deployed with `0.4.x` p-operator and had the stateful sets named using the `pravega-segmentstore` substring.

PR #355, opened 2020-04-04
Issue 354: Segmentstore pods are not coming up with external access enabled

### LINKED ISSUE #354: Segment store pods are not coming up with external access
### Description

Segment store pods are entering in `Crash Loop Back` state with `05.0-rc0` operator with external access eabled
### Importance

Must have
### Location

Operator should be able to bring up segment store pods

anishakj:

### Change log description

In pravega-operator `0.5.0-rc0`, segment store pod name is changed.  And in pravega init scripts we were using pod name to get the details of service. Since pod name has changed, it is not same as service name.   Due to that pravega init scripts are failing.
### Purpose of the change

 Fixes #354

### What the code does

In the case of external access , we will pass `service name` also from pravega-operator. And we have to make changes in pravega scripts to use service name  instead of pod  name.

### How to verify it
Verified that Segment store pods are coming up with external access.

``` kubectl get po
NAME                                                 READY   STATUS    RESTARTS   AGE
bookkeeper-bookie-0                                  1/1     Running   0          24h
bookkeeper-bookie-1                                  1/1     Running   0          24h
bookkeeper-bookie-2                                  1/1     Running   0          24h
bookkeeper-operator-7d4cf4888d-h7mvp                 1/1     Running   0          2d8h
grafana-5cc764cc88-f5ns2                             1/1     Running   0          11d
influxdb-55d54bdc87-m7cqj                            1/1     Running   0          11d
pravega-operator-6b44794589-jglmw                    1/1     Running   0          29h
pravega-pravega-controller-648cff4fc4-d5hbh          1/1     Running   0          21m
pravega-pravega-segment-store-0                      1/1     Running   0          21m
```


pbelgundi:
@anishakj this perhaps needs fixing on segment store startup script as well.
Could you please point to that PR here?

anishakj:
> @anishakj this perhaps needs fixing on segment store startup script as well.
> Could you please point to that PR here?

https://github.com/pravega/pravega/pull/4668

pkg/controller/pravega/pravega_segmentstore.go
```diff
@@ -232,6 +232,9 @@ func MakeSegmentstoreConfigMap(p *api.PravegaCluster) *corev1.ConfigMap {
 
 	if p.Spec.ExternalAccess.Enabled {
 		configData["K8_EXTERNAL_ACCESS"] = "true"
+		serviceName := util.ServiceNameForSegmentStore(p.Name, 0)
+		serviceName = serviceName[0 : len(serviceName)-2]
+		configData["SERVICE_NAME"] = serviceName
```
pbelgundi, line 237: Adding this new variable breaks compatibility of Operator 0.5.0 with older Pravega versions. Why don't we just set env var POD_NAME to point to SSS Servicename instead?

anishakj: we cant set POD_NAME to SSS name, if we do that it in `init-kubernets.sh`  where it uses  `nodename=$( k8 "${ns}" "pods" "${podname}" ".spec.nodeName" )`  fails

anishakj:
Closing this PR as different approach is taken to fix the issue.

PR #362, opened 2020-04-08
Issue 354: Segment store pods are not coming up with external access

### LINKED ISSUE #354: Segment store pods are not coming up with external access

anishakj:

### Change log description

In pravega operator `0.5`, we are creating STS with different name that  in turn changes pod name. Due to that init script of pravega is failing as  it assumes pod name is same service name.  And segment containers are not coming up.POD_NAME is passed as env variable from operator to pravega init script

### Purpose of the change

Fixes #354

### What the code does
From 0.7 pravega onwards, pod name and external service names were different with 0.5 operator.
Changed external service name to pod name, so that external service name and pod name will have the same name in every pravega version. Also while doing upgrade from 0.6 to 0.7, delete the external services and create services with same name of pravega-0.7 segment store pods.

### How to verify it
Tested the below scenarios with external connectivity
1. Able to deploy pravega 0.7 with operator
2. Tried upgarde from 0.6 to 0.7 while IO work load is running. IO has stalled for a minute, while it is upgrading but it is resuming fine
3. Tried Rollback by making the upgrade from 0.6 to 0.7 fail

All the above scenarios have passed

### DIFF
--- a/pkg/controller/pravegacluster/upgrade.go
+++ b/pkg/controller/pravegacluster/upgrade.go
@@ -320,7 +320,6 @@ func (r *ReconcilePravegaCluster) syncControllerVersion(p *pravegav1alpha1.Prave
 }
 
 func (r *ReconcilePravegaCluster) syncSegmentStoreVersion(p *pravegav1alpha1.PravegaCluster) (synced bool, err error) {
-
 	sts := &appsv1.StatefulSet{}
 	name := util.StatefulSetNameForSegmentstore(p)
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, sts)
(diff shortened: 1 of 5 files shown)


pkg/controller/pravegacluster/upgrade.go
```diff
@@ -444,12 +444,29 @@ func (r *ReconcilePravegaCluster) syncSegmentStoreVersionTo07(p *pravegav1alpha1
 	//this check is to see if the newsts is present or not if it's not present it will be created here
 	if err != nil {
 		if errors.IsNotFound(err) {
+			if p.Spec.ExternalAccess.Enabled {
```
pbelgundi, line 447: It would be good to create the services after the new STS is created. Just move this if block  to ln 469

pbelgundi: Encapsulate all the code inside the if block in a method and invoke the same here, instead of adding code inline.

pkg/controller/pravegacluster/upgrade.go
```diff
@@ -564,6 +581,30 @@ func (r *ReconcilePravegaCluster) transitionToNewSTS(p *pravegav1alpha1.PravegaC
 	}
 	//this is to check if all the new ss pods have comeup before deleteing the old sts
 	if newsts.Status.ReadyReplicas == p.Spec.Pravega.SegmentStoreReplicas {
+		if p.Spec.ExternalAccess.Enabled {
```
pbelgundi, line 584: Again, Encapsulate all the code inside the if block in a method and invoke the method here... something like ...
`if p.Spec.ExternalAccess.Enabled { deleteExternalServices(. . .) } `

pkg/util/pravegacluster.go
```diff
@@ -116,6 +119,14 @@ func StatefulSetNameForSegmentstoreBelow07(name string) string {
 	return fmt.Sprintf("%s-pravega-segmentstore", name)
 }
 
+func ServiceNameForSegmentStoreBelow07(name string, index int32) string {
```
pbelgundi, line 122: Do we really need these 2 methods? 
It would be better to have all the name generation logic in **one** place inside method "ServiceNameForSegmentStore" and always invoke this method by passing PravegaCluster object as parameter.

pkg/controller/pravegacluster/upgrade.go
anishakj: > It would be good to create the services after the new STS is created. Just move this if block to ln 469

We wont be able to do that because sts is created with "0" replicas and in pravega.spec.replicas it will set the value., due to that external services wont get created. And we are creating sts in case of external service creation fails.

anishakj: > Encapsulate all the code inside the if block in a method and invoke the same here, instead of adding code inline.

Done

pkg/util/pravegacluster.go
anishakj: Done

Prabhaker24:
LGTM

PR #377, opened 2020-04-29
updated operator.yaml and Readme.md file

Prabhaker24:
### Change log description
changed the image tag in operator.yaml and has changes to the documentation as well

### Purpose of the change
Image tag  is pointing to the latest image of pravega operator on docker hub

### What the code does
when the operator is installed it will download pravega operator image from pravega-operator repo from docker hub

### How to verify it
Install the pravega-operator and pravega cluster


deploy/operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       serviceAccountName: pravega-operator
       containers:
       - name: pravega-operator
-        image: pbelgundi/pravega-operator:m67
+        image: pravega/pravega-operator:0.5.0-rc1
```
pbelgundi, line 20: Use image tag "latest" as that makes sure we don't need to update this file every time we do a new release.

deploy/operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
...
         imagePullPolicy: Always
         ports:
         - containerPort: 6000
```
pbelgundi, line 23: Please also add pravega version "0.7.1" to version-map.yaml .... since it should be out soon...

README.md
```diff
@@ -49,7 +48,11 @@ The Pravega Operator manages Pravega clusters deployed to Kubernetes and automat
 
 > Note: If you are running on Google Kubernetes Engine (GKE), please [check this first](doc/development.md#installation-on-google-kubernetes-engine).
 
-Use Helm to quickly deploy a Pravega operator with the release name `foo`.
+In case you dont have a cert-manager, Install the cert-manager from the following link:-
```
pbelgundi, line 51: We should be saying, "We use cert-manager for certificate management for webhook services in Kubernetes. In case you plan to use the same, you would need to install cert-manager : [link to cert-manager install doc]

deploy/operator.yaml
Prabhaker24: I have changed tag to latest

deploy/operator.yaml
Prabhaker24: i have added pravega version "0.7.1" to the version-map.yaml

README.md
Prabhaker24: i have changed this as well

PR #378, opened 2020-04-29
Updated version_map and charts value

Prabhaker24:

### Change log description
updated version_map and charts value

### Purpose of the change
For supporting the upgrade of pravega to version 0.7.1 entries for the same is added

### What the code does
Allow us to upgrade to pravega version 0.7.1

### How to verify it
Upgrage from version >=0.5.0 to 0.7.1


README.md
```diff
@@ -37,7 +37,7 @@ The Pravega Operator manages Pravega clusters deployed to Kubernetes and automat
 
 ## Requirements
 
-- Kubernetes 1.9+
+- Kubernetes 1.15+
```
pbelgundi, line 40: Thanks for fixing this.

PR #381, opened 2020-04-30
Updating Operator Charts

SrishT:

### Change log description
Updates the operator charts with the required clusterroles.

### Purpose of the change
Assigning the right clusterroles to the operator.

### How to verify it
Using the updated charts you should be able to successfully install pravega operator 0.4.x and a pravega cluster.


SrishT:
@pbelgundi I have tested operator and pravega deployment and deletion using these charts both in default and non-default namespace. Even I/O is running as expected.

Prabhaker24:
LGTM

PR #385, opened 2020-05-04
updated cr.yaml

Prabhaker24:
### Change log description
updated cr-detailed.yaml file and made the cacheVolumeClaimTemplate optional

### Purpose of the change
To provide a sample manifest file 

### What the code does
Now the **cacheVolumeClaimTemplate**  is made optional so there is no need to provide it in case pravega version is >=0.7

### How to verify it
Deploy the manifest file  **example/cr-detailed.yaml**

### DIFF
--- a/example/cr-detailed.yaml
+++ b/example/cr-detailed.yaml
@@ -1,12 +1,12 @@
 apiVersion: "pravega.pravega.io/v1beta1"
 kind: "PravegaCluster"
 metadata:
-  name:  bar-pravega
+  name: "example"
   labels:
     app.kubernetes.io/name: "pravega-cluster"
 spec:
   version: 0.7.0
-  zookeeperUri: pravega-zk-client:2181
+  zookeeperUri: zk-client:2181
 
   # Security configurations for Pravega
   # See https://github.com/pravega/pravega/blob/master/documentation/src/docs/security/pravega-security-configurations.md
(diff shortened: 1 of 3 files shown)


example/cr-detailed.yaml
```diff
@@ -1,85 +1,83 @@
...
         memory: "4Gi"
         cpu: "2000m"
-
-    # Turn on Pravega Debug Logging
+      limits:
+        memory: "8Gi"
+        cpu: "4000m"
     debugLogging: false
-
+    cacheVolumeClaimTemplate:
```
anishakj, line 34: cacheVolumeClaimTemplate is optional for 0.7.0

example/cr-detailed.yaml
```diff
@@ -1,85 +1,83 @@
...
+      storageClassName: standard
+      resources:
+        requests:
+          storage: 20Gi
     longtermStorage:
       filesystem:
         persistentVolumeClaim:
           claimName: pravega-tier2
-
-#      ecs:
```
anishakj, line 48: Can we keep these lines commented, in case someone wants to use ecs for longtermstorage

example/cr-detailed.yaml
Prabhaker24: I have made cache volume template as optional

example/cr-detailed.yaml
Prabhaker24: I have added them back

PR #391, opened 2020-05-07
Issue 390: Segment store pvcs are not getting deleted  even after pravega cluster deletion

### LINKED ISSUE #390: Segment store pvcs are not getting deleted when pravega cluster is deleted
### Description

1.  Install pravega 6.1 using pravega 0.5.0 operator
2. Delete pravega clsuter using `kubectl delete pravegacluster`
3. Segment store pvcs are not getting deleted
```
kubectl get pvc
NAME STATUS VOLUME CAPACITY ACCESS MODES STORAGECLASS AGE
cache-pravega-pravega-segmentstore-0 Bound pvc-3a39e643-269a-4b5a-aa6d-b9db4adc91af 30Gi RWO standard 23s
```

### Importance

should-have

### Location

_(Where is the piece of code, package, or document affected by this issue?)_

_(How do you suggest to fix or proceed with this issue?)_

anishakj:

### Change log description
With the newer controller run time library, setting owner reference to cachevolume template was failing with error `failed to set ownerref: cluster-scoped resource must not have a namespace-scoped owner, owner's namespace default`. Due to which while deleting cluster pvc's are not deleted. Made the changes to create volumeclaim template in cluster namespace, so that owner reference is set correctly.

### Purpose of the change

Fixes #390

### What the code does

Created cache volumeclaims in the same namespace where pravega cluster is created

### How to verify it

Verified that owner reference is set correctly and pvcs are getting deleted while deleting cluster

### DIFF
--- a/pkg/controller/pravega/pravega_segmentstore.go
+++ b/pkg/controller/pravega/pravega_segmentstore.go
@@ -53,7 +53,7 @@ func MakeSegmentStoreStatefulSet(p *api.PravegaCluster) *appsv1.StatefulSet {
 		},
 	}
 	if util.IsVersionBelow07(p.Spec.Version) {
-		statefulSet.Spec.VolumeClaimTemplates = makeCacheVolumeClaimTemplate(p.Spec.Pravega)
+		statefulSet.Spec.VolumeClaimTemplates = makeCacheVolumeClaimTemplate(p)
 	}
 	return statefulSet
 }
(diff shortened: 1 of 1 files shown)


Prabhaker24:
LGTM

pbelgundi:
Good catch, Anisha!
Lets make sure we also fix the same issue on BK and ZK Operators:
https://github.com/pravega/bookkeeper-operator/blob/master/pkg/controller/bookkeepercluster/bookie.go#L198
https://github.com/pravega/zookeeper-operator/blob/6fd02160fbf6a99b515053af47c5958814c23c08/pkg/zk/generators.go#L78

Issue #396, opened 2020-05-15
Including self-validating post-install hooks to the helm charts


SrishT:
### Description
Post-install hooks should be created and should contain all validation necessary to provide Helm a better idea of when the chart is indeed fully installed.

### Importance
nice to have

### Location
pravega helm charts

### Suggestions for an improvement
This change will help in replacing the deployment script with umbrella helm charts.

SrishT:
Fixed with PR https://github.com/pravega/pravega-operator/pull/397

Issue #393, opened 2020-05-15
Add bkLedgersPath parameter to Pravega Helm chart


RaulGracia:
### Description
Currently, deploying Bookkeeper and Pravega via Helm has one usability impediment; the configuration of _Bookkeeper ledgers path_. The problem is that none of the Helm Charts allow us to fully configure the Zookeeper location where the metadata of ledgers is going to be set.

At the moment, the only known way to make this to work is to change the "pravegaCluster" value in the `values.yaml` file of Bookkeeper chart to make it match what Pravega expects later on. But, as one can infer, this is problematic, specially when trying to educate new users on how to deploy a demo Pravega cluster.

While the long term goal may be to have full flexibility in all the parameter settings of Bookkeeper and Pravega via Helm, one immediate fix to this problem would be to extend the Pravega Helm Chart to be able to set the `bkLedgersPath` parameter, which is what the Segment Store uses to lookup for Bookkeeper metadata. With this, we could just execute the default deployment command with Helm for Bookkeeper, and then instruct Pravega where to look for the metadata of Bookkeeper ledgers like this:
```
helm install pravega-operator/charts/pravega --name pravega --set zookeeperUri=zookeeper-zookeeper-client:2181 --set bookkeeperUri="bookkeeper-pravega-bk-bookie-headless.default.svc.cluster.local:3181" --set pravega.longtermStorage=pravega-tier2 --set bkLedgersPath="/bookkeeper/pravega/legers"
```

Concretely, The new parameter to add related to this issue would be:
`--set bkLedgersPath="/bookkeeper/pravega/legers"`

This would make the installation of Pravega and Bookkeeper in Kubernetes much easier, following a natural flow. I think this will be appreciated for new users trying to deploy Pravega in Kubernetes.

### Importance

Should have.

### Location

Pravega Helm Chart.

### Suggestions for an improvement

Add the possibility to set `bkLedgersPath` via Pravega Helm Chart.

SrishT:
Fixed with PR https://github.com/pravega/pravega-operator/pull/397

Issue #304, opened 2019-11-22
Updating Charts to support enabling external access features


SrishT:
### Description

There are a few new external access features for the PravegaCluster that have been added such as support for DNS names. Also annotation support has been added for the Controller and Segmentstore services. But the charts haven't been updated to reflect these changes.

PR #397, opened 2020-05-19
Issue 395: Add helm 3 support, post-install hooks and make cluster specs overridable through helm values

### LINKED ISSUE #395: Making cluster specs overridable by users through Helm values
### Description
A lot of specs in the PravegaCluster and other cluster resources are not overridable in the official cluster charts, and require manual modification or forking. These fields (such as the options or JVM options specs) should be configurable by users  through helm values.

### Importance
must have

### Location
pravega helm charts

SrishT:

### Change log description
Updates the charts and the the helm commands to work with helm v3. It makes all the cluster specs overridable through Helm values. Also adds self validating post-install hooks to the charts.

### Purpose of the change
Fixes #395 #393 #304 #396 

### What the code does
Makes all cluster specs overridable and adds self-validating post-install hooks to the charts.

### How to verify it
User should be able to follow the instructions mentioned in the documentation to install/delete a pravega cluster using charts with helm v3. User should be able to modify any of the pravega cluster specs through the values.yaml file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The Pravega Operator manages Pravega clusters deployed to Kubernetes and automat
 ## Requirements
 
 - Kubernetes 1.15+
-- Helm 2.10+
+- Helm 3+
 - An existing Apache Zookeeper 3.5 cluster. This can be easily deployed using our [Zookeeper operator](https://github.com/pravega/zookeeper-operator)
 - An existing Apache Bookkeeper 4.9.2 cluster. This can be easily deployed using our [BookKeeper Operator](https://github.com/pravega/bookkeeper-operator)
 
@@ -48,13 +48,13 @@ The Pravega Operator manages Pravega clusters deployed to Kubernetes and automat
 
 > Note: If you are running on Google Kubernetes Engine (GKE), please [check this first](doc/development.md#installation-on-google-kubernetes-engine).
 
-We use cert-manager for certificate management for webhook services in Kubernetes. In case you plan to use the same, you would need to install cert-manager : [link to cert-manager install doc]( 
+We use cert-manager for certificate management for webhook services in Kubernetes. In case you plan to use the same, you would need to install cert-manager : [link to cert-manager install doc](
 https://cert-manager.io/docs/installation/kubernetes/)
 
 Then use Helm to quickly deploy a Pravega operator with the release name `foo`.
 
 ```
-$ helm install charts/pravega-operator --name foo
+$ helm install foo charts/pravega-operator
 ```
 
 Verify that the Pravega Operator is running.
@@ -70,7 +70,7 @@ foo-pravega-operator     1         1         1            1           17s
 
 ### Upgrade the Operator
 
-For upgrading the pravega operator refer this link 
+For upgrading the pravega operator refer this link
 https://github.com/pravega/pravega-operator/wiki/Operator-Upgrades
 
 ### Install a sample Pravega cluster
@@ -84,7 +84,7 @@ Check out the available [options for long term storage](doc/longtermstorage.md) 
 For demo purposes, you can quickly install a toy NFS server.
 
 ```
-$ helm install stable/nfs-server-provisioner
+$ helm install stable/nfs-server-provisioner --generate-name
 ```
 
 And create a PVC for longtermStorage that utilizes it.
@@ -98,7 +98,7 @@ $ kubectl create -f ./example/pvc-tier2.yaml
 Use Helm to install a sample Pravega cluster with release name `bar`.
 
 ```
-$ helm install charts/pravega --name bar --set zookeeperUri=[ZOOKEEPER_HOST] --set bookkeeperUri=[BOOKKEEPER_SVC] --set pravega.longtermStorage=[TIER2_NAME]
+$ helm install bar charts/pravega --set zookeeperUri=[ZOOKEEPER_HOST] --set bookkeeperUri=[BOOKKEEPER_SVC] --set storage.longtermStorage.filesystem.pvc=[TIER2_NAME]
 ```
 
 where:
@@ -176,7 +176,7 @@ Check out the [upgrade guide](doc/upgrade-cluster.md).
 ### Uninstall the Pravega cluster
 
 ```
-$ helm delete bar --purge
+$ helm uninstall bar
 $ kubectl delete -f ./example/pvc-tier2.yaml
 ```
 
@@ -185,7 +185,7 @@ $ kubectl delete -f ./example/pvc-tier2.yaml
 > Note that the Pravega clusters managed by the Pravega operator will NOT be deleted even if the operator is uninstalled.
 
 ```
-$ helm delete foo --purge
+$ helm uninstall foo
 ```
 
 If you want to delete the Pravega clusters, make sure to do it before uninstalling the operator.
--- a/charts/pravega-operator/Chart.yaml
+++ b/charts/pravega-operator/Chart.yaml
@@ -1,12 +1,9 @@
+apiVersion: v2
 name: pravega-operator
-version: 0.5.0
-appVersion: 0.5.0
-description: pravega operator Helm chart for Kubernetes
+description: Pravega Operator Helm chart for Kubernetes
+version: 0.1.0
 keywords:
 - pravega
 - storage
-home: https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator
+home: https://github.com/pravega/pravega-operator
 icon: https://avatars3.githubusercontent.com/u/25698199
-sources:
-- https://github.com/pravega/charts/blob/master/pravega-operator
-engine: gotpl
--- a/charts/pravega/Chart.yaml
+++ b/charts/pravega/Chart.yaml
@@ -1,12 +1,9 @@
+apiVersion: v2
 name: pravega
-version: 0.5.0
-appVersion: 0.5.0
-description: pravega Helm chart for Kubernetes
+description: Pravega Helm chart for Kubernetes
+version: 0.1.0
 keywords:
 - pravega
 - storage
 home: https://github.com/pravega/pravega
 icon: https://avatars3.githubusercontent.com/u/25698199
-sources:
-- https://github.com/pravega/pravega
-engine: gotpl
--- a/charts/pravega/README.md
+++ b/charts/pravega/README.md
@@ -8,8 +8,8 @@ This chart creates a [Pravega](https://github.com/pravega/pravega) cluster in [K
 
 ## Prerequisites
 
-  - Kubernetes 1.10+ with Beta APIs
-  - Helm 2.10+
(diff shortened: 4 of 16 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ The Pravega Operator manages Pravega clusters deployed to Kubernetes and automat
 ## Requirements
 
 - Kubernetes 1.15+
-- Helm 2.10+
+- Helm 3+
```
RaulGracia, line 41: Does this mean that Helm 2 is not supported anymore?

bourgeoisor: Not officially I don't think; and it likely shouldn't. Helm2 has some significant security concerns (e.g. Tiller)

This article lays down a couple of the issues with Helm2: https://thenewstack.io/helm-3-is-almost-boring-and-thats-a-great-sign-of-maturity/

Do you see any issues with not supporting Helm2 anymore?

RaulGracia: I asked that because I have a set of instructions to deploy Pravega in Kubernetes and they assume Helm 2, so I will just need to update that. This was the reason for the ask :)

charts/pravega/templates/pravega.yaml
```diff
@@ -5,43 +5,106 @@ metadata:
...
+    {{- if .Values.segmentStore.secret }}
+    segmentStoreSecret:
+      {{- if .Values.segmentStore.secret.name }}
+      secret: {{ .Values.segmentStore.name }}
+      {{- end }}
+      {{- if .Values.segmentStore.secret.path }}
+      mountPath: {{ .Values.segmentStore.path }}
+      {{- end }}
+    {{- end }}
+    controllerjvmOptions: {{ .Values.controller.jvmOptions }}
```
bourgeoisor, line 79: To keep the consistency I would use `JVM` rather than `jvm` in the name of the properties.

bourgeoisor: I believe the JVM properties are looking for lists, which may be harder to give in a single value like that. We may have to do something like this instead:
```
    controllerJVMOptions:
      {{- if .Values.controller.jvmOptions }}
{{ toYaml .Values.controller.jvmOptions | indent 6 }}
      {{- end }}
```

Thoughts?
(and then similar for the other JVM options)

charts/pravega/templates/pravega.yaml
```diff
@@ -5,43 +5,106 @@ metadata:
...
       filesystem:
         persistentVolumeClaim:
-          claimName: {{ .Values.pravega.longtermStorage }}
+          claimName: {{ .Values.storage.longtermStorage.filesystem.pvc }}
+      {{- end }}
+    {{- if .Values.options.enabled }}
+    options:
+      {{- range .Values.options.list }}
+      {{ .key }} : {{ .value | quote }}
+      {{- end }}
```
bourgeoisor, line 109: Might be easier to do:
```
    options:
      {{- if .Values.pravega_options }}
{{ toYaml .Values.pravega_options | indent 6 }}
      {{- end }}
```

That way the values would be a bit simpler to write, e.g.:
```
pravega_options:
  controller.auth.userPasswordFile: "/etc/auth-passwd-volume/password-file.txt"
  controller.auth.tlsEnabled: "false"
  controller.auth.tlsCertFile: "/etc/secret-volume/tls.crt"
  controller.auth.tlsTrustStore: "/etc/secret-volume/tls.crt"
```

charts/pravega-operator/Chart.yaml
```diff
@@ -1,12 +1,9 @@
+apiVersion: v2
 name: pravega-operator
-version: 0.5.0
-appVersion: 0.5.0
-description: pravega operator Helm chart for Kubernetes
+description: Pravega Operator Helm chart for Kubernetes
+version: 0.1.0
```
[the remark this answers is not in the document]
SrishT: Given that this will be the first time we will be publishing our charts, shouldn't we be starting with the version `0.1.0`

charts/pravega/Chart.yaml
```diff
@@ -1,12 +1,9 @@
+apiVersion: v2
 name: pravega
-version: 0.5.0
-appVersion: 0.5.0
-description: pravega Helm chart for Kubernetes
+description: Pravega Helm chart for Kubernetes
+version: 0.1.0
```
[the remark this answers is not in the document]
SrishT: Given that this will be the first time we will be publishing our charts, shouldn't we be starting with the version `0.1.0`

charts/pravega/templates/pravega.yaml
SrishT: The naming convention for `segmentStoreJVMOptions` and `controllerjvmOptions` is not uniform currently. Changing the name to `controllerJVMOptions` would involve code changes and can be done in a separate PR as that does not come under chart changes.

charts/pravega-operator/Chart.yaml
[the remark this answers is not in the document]
bourgeoisor: Fair enough -- just verifying it wasn't an oversight.
What happens when we get to `0.5.0` (which is the current -- albeit unpublished officially -- version)?

(this comment doesn't apply to the other chart of course since _that_ chart was never published unofficially either)

charts/pravega/Chart.yaml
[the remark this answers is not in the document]
bourgeoisor: Fair enough! Resolving this thread

charts/pravega/templates/pravega.yaml
bourgeoisor: Makes sense for the name! The rest of my comment still stands (about using `toYaml` instead of a range over key/value pairs)

charts/pravega/templates/pravega.yaml
SrishT: By specifying options in this way, we get to set the options in the following way :-
```
helm install <name> charts/pravega --set options.enabled=true --set options.list[0].key=bookkeeper.bkLedgerPath --set options.list[0].value="/pravega/bar-pravega/bookkeeper/ledgers"
```
which was a requirement as per #393

charts/pravega/README.md
```diff
@@ -38,15 +38,42 @@ The following table lists the configurable parameters of the Pravega chart and t
...
-| `externalAccess.type` | Pravega external access type | `LoadBalancer` |
-| `bookkeeperUri` | URL for Bookkeeper Servers | `bookkeeper-client:3181` |
-| `pravega.image.repository` | Image repo for Pravega image | `pravega/pravega` |
-| `pravega.controllerReplicas` | Replicas for controller | `1` |
-| `pravega.segmentStoreReplicas` | Replicas for segmentStore | `1` |
-| `pravega.debugLogging` | Enable debug logging | `false` |
-| `pravega.cacheVolumeRequest` | Request storage for cacheVolume | `20Gi` |
-| `pravega.storage.className` | Storage class for Pravega volume | `standard` |
-| `pravega.tier2` | Name of the PVC used for Tier 2 storage | `pravega-tier2` |
+| `version` | Pravega version | `0.5.0` |
```
anishakj, line 41: can we change the pravega version to `0.7.0`

SrishT: done

charts/pravega/templates/post-install-hooks.yaml
```diff
@@ -0,0 +1,119 @@
...
+  verbs:
+  - create
+---
+
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ template "hook.fullname" . }}
+  annotations:
+    "helm.sh/hook": post-install
```
bourgeoisor, line 35: I would suggest changing the hook type to `post-install, post-upgrade`; since what the hook does (check for ready replicas, the right versions, etc) also applies to upgrades. This way there is no need for two different hooks, the same can be used for both.
That comment applies to all the resources here

charts/pravega/templates/post-install-hooks.yaml
```diff
@@ -0,0 +1,119 @@
...
+  resources:
+  - "pods"
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - "pods/exec"
+  verbs:
+  - create
```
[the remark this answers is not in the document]
SrishT: Modified the permissions

charts/pravega/templates/post-install-hooks.yaml
SrishT: Makes sense. Modified the hook type for all resources.

charts/pravega/templates/post-install-hooks.yaml
```diff
@@ -0,0 +1,119 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "hook.fullname" . }}
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+spec:
+  backoffLimit: 3
```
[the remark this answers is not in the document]
SrishT: My thoughts are that we increase the number of retries, the `helm install`  command will keep waiting for that long (30 * number_of_retries) for before giving any output to the user in case there was a failure. Increasing the number of retries might not help in cases where the failure happens due to a generic reason like `ImagePullBackOff`.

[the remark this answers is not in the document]
bourgeoisor: Ok makes sense -- we can always retweak if need-be!

charts/pravega/templates/post-install-upgrade-hooks.yaml
```diff
@@ -0,0 +1,107 @@
...
+      "helm.sh/hook-weight": "1"
+      "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+data:
+  validations.sh: |
+    #!/bin/sh
+    set -e
+    sleep 30
+    kubectl get PravegaCluster -n {{ .Release.Namespace }} {{ template "pravega.fullname" . }} -o jsonpath='{.status.conditions[?(@.type=="PodsReady")].status}' | grep True
+    currentReplicas=`kubectl get PravegaCluster -n {{ .Release.Namespace }} {{ template "pravega.fullname" . }} -o jsonpath='{.status.currentReplicas}'`
+    readyReplicas=`kubectl get PravegaCluster -n {{ .Release.Namespace }} {{ template "pravega.fullname" . }} -o jsonpath='{.status.readyReplicas}'`
```
jkhalack, line 63: Is it possible to have `status.currentReplicas` equal to `status.readyReplicas` but less than `status.replicas`?
I mean the situation where not all the replicas have been created up start, and the ones that have been created are all in the ready state, while some other replicas have not been created yet?

SrishT: I do agree that it would be safer to have `.status.replicas` instead of `.status.currentReplicas`. I have incorporated that change.

PR #398, opened 2020-05-28
Add certs overrides and improve chart resource metadata

bourgeoisor:
### Change log description

This PR adds overrides to the Pravega Operator webhook certs as well as the ability to use self-signed or not. Metadata throughout the chart is also enhanced with explicit namespaces.

### Purpose of the change

These changes are needed to support non-self-signed custom certificates as well as generating the resources using `helm template`.

### What the code does

Listed below are all the changes done in the Pravega Operator charts:
- Added explicit `{{ .Release.Namespace }}` namespaces in all namespace-bound resources (to support helm template)
- Moved all certificate related resources into the same file (for readability)
- Added BookkeeperCluster permissions to the operator roles (to support upgrades)
- Added standard best-practice Helm labels (for ease of CLI commands)
- Modified the `fullname` template to allow for full override
- Fully support non-default namespaces by interpolating the release namespace where it was using `default`
- Add support for external non-self-signed certificates

### How to verify it

We can use `helm template` to verify that all the changes are occurring as they are supposed to. We can see among other things the `metadata.namespace` being templated in, as well as the best-practice Helm labels correctly generating.

#### With self-signed certs:
`values.yaml`:
```
webhookCert:
  selfsigned: true
```

Notice that no custom Secret is taken and that the Issuer is to create its own self-signed certificate. We can also notice the Certificate using the correct namespaces instead of `default`.
```
# Source: pravega-operator/templates/certificate.yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Issuer
metadata:
  name: selfsigned-cert
  namespace: pravega
  labels:
    app.kubernetes.io/name: pravega-operator
    app.kubernetes.io/version: "0.5.0"
    helm.sh/chart: "pravega-operator-0.5.0"
spec:
  selfSigned: {}
---
# Source: pravega-operator/templates/certificate.yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
metadata:
  name: selfsigned-cert
  namespace: pravega
  labels:
    app.kubernetes.io/name: pravega-operator
    app.kubernetes.io/version: "0.5.0"
    helm.sh/chart: "pravega-operator-0.5.0"
spec:
  isCA: true
  duration: 8760h
  renewBefore: 360h
  secretName: selfsigned-cert-tls
  commonName: pravega-webhook-svc.pravega.svc.cluster.local
  dnsNames:
    - pravega-webhook-svc
    - pravega-webhook-svc.pravega.svc.cluster.local
    - pravega-webhook-svc.pravega.svc
  issuerRef:
    name: selfsigned-cert
```

#### With custom certs:
`values.yaml`:
```
webhookCert:
  selfsigned: false
  crt: fake_cert_crt_LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURyRENDQX
  key: fake_cert_key_LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURyRENDQX
```

Notice that a Secret for the custom certificate Secret gets created and that the Issuer takes that Secret instead of generating its own. We can also notice the Certificate using the correct namespaces instead of `default`.
```
# Source: pravega-operator/templates/certificate.yaml
apiVersion: v1
kind: Secret
metadata:
  name: pravega-operator-key-pair
  namespace: pravega
type: kubernetes.io/tls
data:
  tls.crt: fake_cert_crt_LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURyRENDQX
  tls.key: fake_cert_key_LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURyRENDQX
---
# Source: pravega-operator/templates/certificate.yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Issuer
metadata:
  name: selfsigned-cert
  namespace: pravega
  labels:
    app.kubernetes.io/name: pravega-operator
    app.kubernetes.io/version: "0.5.0"
    helm.sh/chart: "pravega-operator-0.5.0"
spec:
  ca:
    secretName: pravega-operator-key-pair
---
# Source: pravega-operator/templates/certificate.yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
metadata:
  name: selfsigned-cert
  namespace: pravega
  labels:
    app.kubernetes.io/name: pravega-operator
    app.kubernetes.io/version: "0.5.0"
    helm.sh/chart: "pravega-operator-0.5.0"
spec:
  isCA: true
  duration: 8760h
  renewBefore: 360h
  secretName: selfsigned-cert-tls
  commonName: pravega-webhook-svc.pravega.svc.cluster.local
  dnsNames:
    - pravega-webhook-svc
    - pravega-webhook-svc.pravega.svc.cluster.local
    - pravega-webhook-svc.pravega.svc
  issuerRef:
    name: selfsigned-cert
```

#### Testing with PravegaCluster

Additionally, we can deploy a full PravegaCluster to verify that it deploys without any errors:
```
pravega   pravega-controller-74865546f-z6csr         1/1     Running     0          62m
pravega   pravega-grafana-0                          2/2     Running     0          62m
pravega   pravega-influxdb-0                         1/1     Running     0          62m
pravega   pravega-segment-store-0                    1/1     Running     0          62m
pravega   pravega-operator-6557bf68c6-whw5g          1/1     Running     0          67m
```


charts/pravega-operator/templates/certificate.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Issuer
+metadata:
+  name: selfsigned-cert
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: pravega-operator
```
SrishT, line 7: shouldn't this be `{{- define "pravega-operator.fullname" -}}` instead of `pravega-operator`

charts/pravega-operator/templates/certificate.yaml
```diff
@@ -0,0 +1,50 @@
...
+    app.kubernetes.io/version: "{{ .Chart.AppVersion }}"
```
SrishT, line 8: the appVersion has been removed from Chart.yaml since it requires separate maintenance. In that case the version can be picked from `image.tag`. What do u think?

charts/pravega-operator/templates/certificate.yaml
```diff
@@ -0,0 +1,50 @@
...
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+spec:
+{{- if .Values.webhookCert.selfsigned }}
+  selfSigned: {}
+{{- else }}
+  ca:
+    secretName: pravega-operator-key-pair
```
SrishT, line 15: the names for all resources are generated in `_helpers.tpl` where the release name is prepended to the resource name.Could we follow that approach for this one well to ensure consistency?

charts/pravega-operator/templates/webhook.yaml
```diff
@@ -2,7 +2,12 @@ apiVersion: v1
 kind: Service
 metadata:
   name: pravega-webhook-svc
+  namespace: {{ .Release.Namespace }}
   namespace: default
```
SrishT, line 6: This line needs to be removed

charts/pravega-operator/values.yaml
```diff
@@ -1,29 +1,35 @@
...
 
-# Whether to create custom resource
+## Whether to create the CRD.
 crd:
   create: true
 
-# whether to enable test mode
-testmode: false
+  ## Whether to enable test mode.
+  testmode: false
```
jkhalack, line 24: Extra indentation?

charts/pravega-operator/templates/certificate.yaml
bourgeoisor: Out of curiosity, is it much more work to change the string in both the `appVersion` and `image.tag` when we do versioning bumps in this repo?

From what I can see, `app.kubernetes.io/version: "{{ .Chart.AppVersion }}"` is the best practice: https://helm.sh/docs/chart_best_practices/labels/ (see the "Standard Labels" section)

bourgeoisor: `appVersion` is also what appears when you do `helm list`

charts/pravega-operator/templates/_helpers.tpl
```diff
@@ -9,8 +9,23 @@ Expand the name of the chart.
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
```
bourgeoisor, line 12: Modified this to allow for a `fullnameOverride` as well, keeping the same pattern as other big OSS projects such as etcd: https://github.com/helm/charts/blob/master/incubator/etcd/templates/_helpers.tpl#L9-L25
Without an override by default this will act just like before.

bourgeoisor:
Resolved all the comments (except the `appVersion` one, still in-discussion), fixed the merge conflicts, and re-pushed.

charts/pravega-operator/templates/_helpers.tpl
```diff
@@ -9,8 +9,23 @@ Expand the name of the chart.
...
 */}}
 {{- define "pravega-operator.fullname" -}}
+{{- if .Values.fullnameOverride -}}
```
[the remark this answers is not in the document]
bourgeoisor: Nope, no need! It's only used _if_ it exists, else `pravega-operator.fullname` will just be what it used to be before.

charts/pravega-operator/templates/crd.yaml
```diff
@@ -38,13 +38,23 @@ spec:
...
+       {{- end }}
        service:
           name: pravega-webhook-svc
-          namespace: default
+          namespace: {{ .Release.Namespace }}
           path: /convert
+  version: v1beta1        
   versions:
   - name: v1beta1
+    {{- if .Release.IsUpgrade }}
```
anishakj, line 53: we have to make this change for version v1alpha1 not v1beta1

charts/pravega-operator/templates/certificate.yaml
```diff
@@ -0,0 +1,46 @@
...
+metadata:
+  name: {{ template "pravega-operator.fullname" . }}-selfsigned-cert
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "pravega-operator.commonLabels" . | indent 4 }}
+spec:
+  isCA: true
+  duration: 8760h
+  renewBefore: 360h
+  secretName: {{ template "pravega-operator.fullname" . }}-selfsigned-cert-tls
```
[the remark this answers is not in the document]
SrishT: @jkhalack @bourgeoisor do u think truncating the `fullname` to 40 characters would be a good way to go about it?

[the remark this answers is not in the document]
bourgeoisor: It could definitely help -- but probably outside of the scope of this PR.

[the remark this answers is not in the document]
SrishT: Agreed. Maybe this could be taken up in a separate issue.

PR #403, opened 2020-06-15
Issue 402: Support for name override inside charts

### LINKED ISSUE #402: Adding name overrides for pravega cluster name through charts
### Description
In the charts we need to give the user the provision of overriding the pravega cluster name (which is defaulted to `<release-name>-<chart-name>`) by specifying the desired value for `fullNameOverride` in the values.yaml

### Importance
nice-to-have

### Location
charts

SrishT:

### Change log description
Gives the provision to override the pravega cluster name by providing the value for `fullNameOverride` inside values.yaml

### Purpose of the change
Fixes #402 

### What the code does
- If release name is same as the chart name, chart name is used as the resource name.
- If `fullNameOverride` is provided inside the values.yaml file, that value is used as the resource name.

### How to verify it
Tested 3 different scenarios for naming the resources
- `fullNameOverride` is mentioned in values.yaml file (resource name uses the value provided for this field)
- `fullNameOverride` is not mentioned, release name same as chart name (resource name is same as the chart name)
- `fullNameOverride` is not mentioned, release name is different from chart name (resource name is in the given format `<release-name>-<chart-name>`)


charts/pravega-operator/templates/certificate.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     secretName: {{ template "pravega-operator.fullname" . }}-key-pair
 {{- end }}
 ---
-apiVersion: certmanager.k8s.io/v1alpha1
+apiVersion: cert-manager.io/v1alpha2
 kind: Certificate
 metadata:
   name: {{ template "pravega-operator.fullname" . }}-selfsigned-cert
```
anishakj, line 19: Are we planning to change pravega-operator full name , so that by adding `selfsigned-cert-tls` it should not exceed 63 characters

SrishT: The possibility for this scenario to occur is rare, and if the need is felt, will make this change for all the charts, but for now I think this should work

charts/pravega-operator/templates/certificate.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: certmanager.k8s.io/v1alpha1
+apiVersion: cert-manager.io/v1alpha2
```
jkhalack, line 1: What is the reason for `apiVersion` change?

PR #405, opened 2020-06-15
Issue 404 : Added Readiness check for controller

### LINKED ISSUE #404: Pravega-operator marks pravega deployment as Ready before pravega is operational
### Description
Pravega-operator marks pravega deployment as Ready before pravega is operational

### Importance
Must_have

### Location
we need to add a readiness probe to insure controller is ready state before pravega-operator marks pravega deployment as ready.

Add Readiness probe check.

Prabhaker24:

### Change log description
Pravega-operator marks pravega deployment as Ready before pravega is operational so a readiness check is needed to ensure pravega deployment doesn't mark pravega ready before it actually is.

### Purpose of the change
This pr fixes #404 

### What the code does
This code introduce readiness check for the controller pod and will only mark controller ready when it's actually ready and is able to create scopes and streams.

### How to verify it
1) To verify this I have deployed pravega and it marked controller in ready state only when the 
 controller is able to create stream and scopes.

2) I have also verified by deploying pravega where the controller was not able to create scope and 
 in that case, pravega deployment didn't mark the deployment ready.

### DIFF
--- a/pkg/controller/pravega/pravega_controller.go
+++ b/pkg/controller/pravega/pravega_controller.go
@@ -100,7 +100,7 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
 							Command: util.ControllerReadinessCheck(10080),
 						},
 					},
-					// Controller pods start fast. We give it up to 60 seconds to become ready.
+					// Controller pods start fast. We give it up to 20 seconds to become ready.
 					InitialDelaySeconds: 20,
 					TimeoutSeconds:      60,
 					SuccessThreshold:    3,
--- a/pkg/controller/pravega/pravega_controller.go
+++ b/pkg/controller/pravega/pravega_controller.go
@@ -97,12 +97,13 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
 				ReadinessProbe: &corev1.Probe{
 					Handler: corev1.Handler{
 						Exec: &corev1.ExecAction{
-							Command: util.ReadinessHealthcheckCommand(10080),
+							Command: util.ControllerReadinessCheck(10080),
 						},
 					},
 					// Controller pods start fast. We give it up to 1 minute to become ready.
-					PeriodSeconds:    5,
-					FailureThreshold: 12,
+					InitialDelaySeconds: 60,
+					TimeoutSeconds:      10,
+					SuccessThreshold:    3,
 				},
 				LivenessProbe: &corev1.Probe{
 					Handler: corev1.Handler{
@@ -130,7 +131,6 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
 			},
 		},
 	}
-
 	if p.Spec.Pravega.ControllerServiceAccountName != "" {
 		podSpec.ServiceAccountName = p.Spec.Pravega.ControllerServiceAccountName
 	}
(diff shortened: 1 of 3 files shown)


pkg/util/pravegacluster.go
```diff
@@ -64,6 +64,10 @@ func HealthcheckCommand(port int32) []string {
 	return []string{"/bin/sh", "-c", fmt.Sprintf("netstat -ltn 2> /dev/null | grep %d || ss -ltn 2> /dev/null | grep %d", port, port)}
 }
 
+func ReadinessHealthcheckCommand(port int32) []string {
```
pbelgundi, line 67: Would be good to call this method `ControllerReadinessCheck()` to indicate its appropriate purpose....

pkg/controller/pravega/pravega_controller.go
```diff
@@ -97,7 +97,7 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
 				ReadinessProbe: &corev1.Probe{
 					Handler: corev1.Handler{
 						Exec: &corev1.ExecAction{
-							Command: util.HealthcheckCommand(9090),
+							Command: util.ReadinessHealthcheckCommand(10080),
```
pbelgundi, line 100: In the ReadinessProbe configuration, we seem to be setting only "periodSeconds" and "failureThreshold".... These values need to be adjusted as per the new check....

 "failureThresold" value is very high... perhaps required by the earlier check... 
but now based on the new check, it can be removed ( _so default of 3 would be used_).

Based on K8s documentation for probes:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#probe-v1-core
it may be good to set "initialDelaySeconds" to avoid doing the probe till the controller server has had time to come up successfully...
Also consider setting "timeoutSeconds" to guard from situations where curl hangs...
Also consider setting "successThreshold" ...

pkg/util/pravegacluster_test.go
```diff
@@ -175,6 +175,12 @@ var _ = Describe("pravegacluster", func() {
 		})
 
 	})
+	Context("ReadinessHealthcheckCommand()", func() {
```
pbelgundi, line 178: e2e tests are failing on travis... please fix if needed

pkg/util/pravegacluster.go
Prabhaker24: I have changed it.

pkg/controller/pravega/pravega_controller.go
Prabhaker24: yes, now I have removed "failureThresold" and "periodSeconds" as there default values are enough and have added the other 3 fields.

pkg/util/pravegacluster_test.go
Prabhaker24: yes, they were getting timeout on restarting them again it passed.

pkg/util/pravegacluster.go
pbelgundi: ok.

pkg/controller/pravega/pravega_controller.go
pbelgundi: Ok

pkg/controller/pravega/pravega_controller.go
```diff
@@ -97,12 +97,13 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
...
 						Exec: &corev1.ExecAction{
-							Command: util.HealthcheckCommand(9090),
+							Command: util.ControllerReadinessCheck(10080),
 						},
 					},
 					// Controller pods start fast. We give it up to 1 minute to become ready.
-					PeriodSeconds:    5,
-					FailureThreshold: 12,
+					InitialDelaySeconds: 20,
+					TimeoutSeconds:      10,
```
pbelgundi, line 105: timeout of 10 secs for curl is very low... should be about 60 secs...

Prabhaker24: Done.

PR #409, opened 2020-06-25
Issue 388: Adding support for helm upgrades

### LINKED ISSUE #388: Add support for versioning and publishing of helm charts for supporting helm upgrades
### Description
Currently our helm charts are not versioned or published.
There is no conscious effort to maintain upgrade compatibility in charts.

### Importance
must-have

### Location
https://github.com/pravega/pravega-operator/tree/master/charts

SrishT:

### Change log description
To add support to upgrade pravega operator from version `0.4.x` to `0.5.0` via helm.

### Purpose of the change
Fixes #388 

### How to verify it
The user needs to follow the given steps in order to trigger the upgrade
- Install bookkeeper operator and cert-manager
- Install the issuer and certificate using the manifest `deploy/certificate.yaml` and plug the value `tls.crt` from the secret created by the certificate that has been created in this step using the following command
```
$ kubectl get secret selfsigned-cert-tls -o yaml | grep tls.crt
```
Here `selfsigned-cert-tls` is the name of the secret used in the file `deploy/certificate.yaml`
The `tls.crt` value so obtained needs to be provided either in the `webhookCert.crt` field of the values.yaml file of the 0.5.0 operator charts or in the `--set webhookCert.crt=<tls.crt>`  of the helm upgrade command.
- Execute the script `./scripts/pre-upgrade.sh <operator release name> <operator release namespace>`
- Trigger the operator upgrade. To trigger the upgrade, run the following command
```
$ helm upgrade <operator release name> <location of 0.5.0 charts> --set webhookCert.crt=<tls.crt>
```

You will see the following message as soon as the operator upgrade is done
```
Release <operator release name> has been upgraded. Happy Helming!
NAME: <operator release name>
LAST DEPLOYED: Thu Jun 25 14:21:02 2020
NAMESPACE: <operator release namespace>
STATUS: deployed
REVISION: 2
TEST SUITE: None
```

Soon after this, the conversion of the Pravega CR from from `v1alpha1` to `v1beta1` is triggered and the ownership of the Bookkeeper CR is migrated from the Pravega Operator to the Bookkeeper Operator.
The presence of the following log message in the Pravega Operator logs indicates the successful completion of the Pravega Operator version migration.
```
Version migration completed successfully
```


SrishT:
@Prabhaker24 I will be modifying the documentation once the charts have been published since this approach will be usable completely only once that has been done too.

pbelgundi:
> @Prabhaker24 I will be modifying the documentation once the charts have been published since this approach will be usable completely only once that has been done too.

@SrishT Documentation should be updated in this PR based on code changes in this PR. Once charts are published we can change docs again if needed.

doc/operator-upgrade.md
```diff
@@ -82,18 +82,26 @@ For upgrading Operator to version 0.5.0, the following must be true:
...
+4. Install an Issuer and a Certificate (either self-signed or CA signed) in the same namespace as the Pravega Operator (refer to [this](https://github.com/pravega/pravega-operator/blob/master/deploy/certificate.yaml) manifest to create a self-signed certificate in the default namespace).
+> The name of the certificate (*webhookCert.certName*), the name of the secret created by this certificate (*webhookCert.secretName*), the tls.crt (*webhookCert.crt*) and tls.key (*webhookCert.key*) need to be specified against the corresponding fields in the values.yaml file, or can be provided with the upgrade command as shown [here](#triggering-the-upgrade).
+The values *tls.crt* and *tls.key* are contained in the secret which is created by the certificate and can be obtained using the following command
+```
+kubectl get secret <secret-name> -o yaml | grep tls.
+```
+
+5. Execute the script `pre-upgrade.sh` inside the `scripts` folder. This script patches the `pravega-webhook-svc` with the required annotations and labels.
```
Prabhaker24, line 98: We have the upgrade script present in the tools folder a better option, in my opinion, would be to not remove the documentation part on how to use it rather we can just say a better or recommended way of upgrading operator from 0.4.x to 0.5.0 is using helm and mention the steps for it.

SrishT: The script still exists but we should not be mentioning both the approaches in the same document is what I feel. Hence even though the script is still there we should still mention just the helm approach here. Both approaches mentioned in the same place would just increase confusion.

doc/upgrade-cluster.md
```diff
@@ -23,29 +23,39 @@ Your Pravega cluster should be in a healthy state. You can check your cluster he
 ```
 $ kubectl get PravegaCluster
 NAME      VERSION   DESIRED MEMBERS   READY MEMBERS   AGE
-example   0.4.0     7                 7               11m
+pravega   0.4.0     7                 7               11m
 ```
 
 ## Upgrade Path Matrix
```
pbelgundi, line 30: Now that this is moved to a config map, lets remove this version-map from the documentation. Instead we could point to configmap.yaml

scripts/pre-upgrade.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
```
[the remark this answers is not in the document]
SrishT: The script needs to be invoked manually even when we are doing helm upgrade since the helm `pre-upgrade hooks` are executed on an upgrade request after templates are rendered, but before any resources are updated.

doc/upgrade-cluster.md
SrishT: done

scripts/pre-upgrade.sh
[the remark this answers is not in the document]
pbelgundi: Ok.

PR #427, opened 2020-07-20
Pravega Operator API Resource Hook.

co-jo:
### Change log description

* Adds a manifest which defines a post-install/upgrade hook to make sure the `pravegacluster` resource exists.

### Purpose of the change

* If one is unlucky with the timings when install the `pravega-operator` chart, directly followed by the `pravega` chart, the api-resource `PravegaCluster` may not exist by the time the pravega chart tries to create a `PravegaCluster`.

### What the code does

* See above.

### How to verify it

Run many install/uninstall cycles of the mentioned charts -- see that the pravega chart does not start before its resource dependency is registered.

### DIFF
--- a/charts/pravega-operator/templates/post-install-upgrade-hooks.yaml
+++ b/charts/pravega-operator/templates/post-install-upgrade-hooks.yaml
@@ -72,15 +72,6 @@ data:
     if [ -z "$(kubectl api-resources | grep PravegaCluster)" ]; then
         exit 1
     fi
-
-    readyReplicas=`kubectl get deployment {{ template "pravega-operator.fullname" . }} \
-      -n {{ .Release.Namespace }} \
-      -o jsonpath='{.status.readyReplicas}'`
-    echo "PravegaCluster readyReplicas: $readyReplicas"
-
-    if [ -z $readyReplicas ] || [ $readyReplicas -le 0 ]; then
-        exit 1
-    fi
 ---
 
 apiVersion: batch/v1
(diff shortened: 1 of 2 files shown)


charts/pravega-operator/templates/post-install-upgrade-hooks.yaml
```diff
@@ -0,0 +1,107 @@
...
+kind: Job
+metadata:
+  name: {{ template "pravega-operator.fullname" . }}-post-install-upgrade
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-weight": "2"
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation, hook-failed
+spec:
+  backoffLimit: 3
```
SrishT, line 87: Could u use the backOffLimit value that u have taken as input from the user in values.yaml instead of hardcoding it?

charts/pravega-operator/values.yaml
```diff
@@ -40,3 +40,9 @@ webhookCert:
 ## Specifies which namespace the Operator should watch over.
 ## An empty string means all namespaces.
 watchNamespace: ""
+
+hooks:
+  image:
+    repository: lachlanevenson/k8s-kubectl
+    tag: v1.16.10
+  backoffLimit: 10
```
SrishT, line 48: This is the value u should be using.

Issue #357, opened 2020-04-06
Change health(readiness) checks for Controller and SegmentStore


pbelgundi:
### Description

Current readiness checks for Controller and SegmentStore are only placeholders as the same simple port check is used for Liveness and Readiness Probes.
Readiness Probes need to be more application specific checks thaat enable us to determine if a server is capable of serving more requests or not.

### Importance

should-have

### Location

https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/pravega_controller.go#L100
https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/pravega_segmentstore.go#L126

### Suggestions for an improvement
Depends on https://github.com/pravega/pravega/issues/1665

pbelgundi:
This is blocked till probes are implemented on Pravega side

PR #433, opened 2020-07-29
Issue 432: Controller Readiness probe is failing in case of ingress

### LINKED ISSUE #432: Controller Readiness probe is failing in case of ingress
### Description
The controller Readiness probe is failing in case of ingress

### Importance
Must_Have

### Location
This requires changes in pravega_controller.go and pravegacluster.go file

Prabhaker24:
### Change log description
 The controller Readiness probe is failing in case of ingress

### Purpose of the change
fixes #432 

### What the code does
The code here does a readiness check for controller in the following 4 cases 
1) Auth and TLS Enabled- in this case we check for the controller is properly enabled with authentication or not and we do a get on controller and with Dummy credentials(testtls:testtls) and the controller returns 401 error in this case if it's correctly configured
2) Auth Enabled and TLS Disabled- in this case, we check for the controller is properly enabled with authentication or not and we do a get on controller and with Dummy credentials(testtls:testtls) and the controller returns 401 error in this case if it's correctly configured
3) Auth Disabled and Tls Enabled- in this case we check if the controller is able to create scopes or not by checking if system scope is present or not
4) Auth and TLs Disabled- in this case we check if the controller is able to create scopes or not by checking if system scope is present or not

### How to verify it
To verify try and check for all the above 4 cases and see if the controller pod is marked as ready or not post-deployment

### TODO
For the case of Auth, when we get an API at the pravega side which would do the check for the presence of system scope then we can probably call that in our readiness probe to solve the case of controller pod being marked ready even though it can't create scope and stream.

### DIFF
--- a/pkg/util/pravegacluster.go
+++ b/pkg/util/pravegacluster.go
@@ -65,10 +65,19 @@ func HealthcheckCommand(port int32) []string {
 }
 
 func ControllerReadinessCheck(port int32, authflag bool) []string {
+	// This is to check the readiness of controller in case auth is Enabled
+	// here we are using login credential as testtls:testtls which should
+	// not be used as auth credential and we depend on controller giving us
+	// 401 error which means controller is properly configured with auth
+	// it checks both cases when tls is enabled as well as tls dissabled
+	// with auth enabled
 	if authflag == true {
 		return []string{"/bin/sh", "-c", fmt.Sprintf("echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=true' &&  curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 || (echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=false' && curl -v -k -u testtls:testtls -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 ) ||  (echo $JAVA_OPTS | grep 'controller.security.tls.enable=true' && echo $JAVA_OPTS | grep -v 'controller.auth.tlsEnabled' && curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 ) || (curl -v -k -u testtls:testtls -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 )", port, port, port, port)}
 	}
-	return []string{"/bin/sh", "-c", fmt.Sprintf("echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=true' &&  curl -s -X GET 'http://localhost:%d/v1/scopes/' -H 'accept: application/json' | grep '_system'|| (echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=false' && curl -s -X GET 'http://localhost:%d/v1/scopes/' -H 'accept: application/json' | grep '_system' ) || (echo $JAVA_OPTS | grep 'controller.security.tls.enable=true' && echo $JAVA_OPTS | grep -v 'controller.auth.tlsEnabled' && curl -s -X GET 'http://localhost:%d/v1/scopes/' -H 'accept: application/json' | grep '_system' ) || (curl -s -X GET 'http://localhost:%d/v1/scopes/' -H 'accept: application/json' | grep '_system') ", port, port, port, port)}
+	// This is to check the readiness in case auth is not enabled
+	// and it covers both the cases with tls enabled and tls dissabled
+	// along with auth dissabled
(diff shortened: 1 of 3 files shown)


pkg/util/pravegacluster.go
```diff
@@ -64,17 +64,20 @@ func HealthcheckCommand(port int32) []string {
...
+		return []string{"/bin/sh", "-c", fmt.Sprintf("echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=true' &&  curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 || (echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=false' && curl -v -k -u testtls:testtls -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 ) ||  (echo $JAVA_OPTS | grep 'controller.security.tls.enable=true' && echo $JAVA_OPTS | grep -v 'controller.auth.tlsEnabled' && curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 ) || (curl -v -k -u testtls:testtls -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 )", port, port, port, port)}
 	}
-	return []string{"/bin/sh", "-c", fmt.Sprintf("curl -s -X GET 'http://localhost:%d/v1/scopes/' -H 'accept: application/json' | grep '_system'", port)}
+	// This is to check the readiness in case auth is not enabled
+	// and it covers both the cases with tls enabled and tls dissabled
```
anishakj, line 78: change to disabled

Prabhaker24: Done.

pkg/util/pravegacluster.go
```diff
@@ -64,17 +64,20 @@ func HealthcheckCommand(port int32) []string {
...
+	// along with auth dissabled
```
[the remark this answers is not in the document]
Prabhaker24: Done.

pkg/util/pravegacluster.go
```diff
@@ -64,17 +64,20 @@ func HealthcheckCommand(port int32) []string {
 	return []string{"/bin/sh", "-c", fmt.Sprintf("netstat -ltn 2> /dev/null | grep %d || ss -ltn 2> /dev/null | grep %d", port, port)}
 }
 
-func ControllerReadinessCheck(port int32, authflag bool, tlsflag bool) []string {
-	if authflag == true && tlsflag == true {
-		return []string{"/bin/sh", "-c", fmt.Sprintf("curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401", port)}
-	}
+func ControllerReadinessCheck(port int32, authflag bool) []string {
```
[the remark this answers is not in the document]
Prabhaker24: Done.

pkg/util/pravegacluster.go
```diff
@@ -64,17 +64,25 @@ func HealthcheckCommand(port int32) []string {
...
-		return []string{"/bin/sh", "-c", fmt.Sprintf("curl -v -k -u testauth:testauth -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401", port)}
-	}
-	if tlsflag == true {
-		return []string{"/bin/sh", "-c", fmt.Sprintf("curl -v -k -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep '_system'", port)}
+		return []string{"/bin/sh", "-c", fmt.Sprintf("echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=true' &&  curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 || (echo $JAVA_OPTS | grep 'controller.auth.tlsEnabled=false' && curl -v -k -u testtls:testtls -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 ) ||  (echo $JAVA_OPTS | grep 'controller.security.tls.enable=true' && echo $JAVA_OPTS | grep -v 'controller.auth.tlsEnabled' && curl -v -k -u testtls:testtls -s -X GET 'https://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 ) || (curl -v -k -u testtls:testtls -s -X GET 'http://localhost:%d/v1/scopes/' 2>&1 -H 'accept: application/json' | grep 401 )", port, port, port, port)}
```
sarlaccpit, line 80: I understand the intent about inducing 401 errors to validate that the rest api is responsive and secured but...  the big downside is that it is flooding the controller logs with authentication errors that are very very confusing and making it appear like there are some authentication issues in the system.

Wondering if there's a better way to achieve this....

Prabhaker24: Currently, this is the best we could do from the operator side and this is a temporary fix which we will change once we have api's that we can call for doing health check at the pravega side link to the pravega issue is https://github.com/pravega/pravega-operator/issues/357

sarlaccpit: Understood thank you.

PR #440, opened 2020-08-18
Issue 439: Post upgrade script for helm upgrades

### LINKED ISSUE #439: Pravega and Bookkeeper releases not updated after upgrading pravega operator from 0.4 to 0.5 via helm
### Description
After pravega operator is upgraded from 0.4.5 to 0.5.x via helm, the helm release corresponding to the `PravegaCluster` resource is not updated and a separate helm release is not created corresponding to the newly created `BookkeeperCluster` resource. This prevents us from being able to do any further helm operations on the pravega and bookkeeper clusters post the operator upgrade.

### Importance
must have

SrishT:

### Change log description
After upgrading pravega operator from 0.4.5 to 0.5.x via helm, the helm release corresponding to the `PravegaCluster` resource is not updated and a separate helm release is not created corresponding to the newly created `BookkeeperCluster` resource. Due to this, no further helm operations can be done on the pravega and bookkeeper clusters post the operator upgrade.

### Purpose of the change
Fixes #439 

### What the code does
Upgrades the helm release corresponding to the `PravegaCluster` resource and creates a separate helm release corresponding to the newly created `BookkeeperCluster` resource, after pravega operator is upgraded from 0.4.5 to 0.5.x via helm.

### How to verify it
After the operator upgrade completes, run the post-upgrade script in the following way
```
$ ./post-upgrade.sh <name of pravega cluster> <pravega release name> <bookkeeper release name> <version of pravega cluster> <namespace> <zookeeper svc name>
```
After this script executes successfully, helm upgrade and helm uninstall on the pravega and bookkeeper releases will also succeed.


doc/operator-upgrade.md
```diff
@@ -115,6 +115,18 @@ The upgrade to Operator 0.5.0 can be triggered using the following command
 helm upgrade <operator release name> <location of 0.5.0 charts> --set webhookCert.crt=<tls.crt> --set webhookCert.generate=false --set webhookCert.certName=<cert-name> --set webhookCert.secretName=<secret-name>
```
bourgeoisor, line 115: Should it be obvious here that the `helm upgrade` command should be run three times (once for each of the operators), or are those instructions in each operator's GitHub repo?

doc/operator-upgrade.md
```diff
@@ -115,6 +115,18 @@ The upgrade to Operator 0.5.0 can be triggered using the following command
...
 ```
 
+Once the upgrade completes and the following command starts returning the expected response
```
bourgeoisor, line 118: Could be useful to know what "the expected response" was (i.e. what to look for exactly)

doc/operator-upgrade.md
```diff
@@ -115,6 +115,18 @@ The upgrade to Operator 0.5.0 can be triggered using the following command
...
+```
+kubectl describe PravegaCluster
+```
+Execute the script `post-upgrade.sh` inside the [scripts](https://github.com/pravega/pravega-operator/blob/master/scripts) folder. This script patches the `PravegaCluster` and `BookkeeperCluster` resources with the required annotations and labels, and updates their corresponding helm releases. This script needs the following arguments
```
bourgeoisor, line 122: Perhaps an example of what a full command looks like (e.g. `./post-upgrade.sh <PravegaCluster resource name> <Pravega cluster Helm release>......`)

bourgeoisor: Side-note: the less argument, the better / simpler it is for the users. If there are arguments that could be trimmed, removed, combined, that'd be good for them

What about optional flags, where if you don't put them we used some sensible default? Harder with Bash of course but _if_ Python is acceptable that's an alternative.

Just throwing some ideas in the air

scripts/post-upgrade.sh
```diff
@@ -0,0 +1,46 @@
...
+	exit 1
+fi
+
+name=$1
+pname=$2
+bkname=$3
+version=$4
+namespace=$5
+zksvc=$6
+
```
bourgeoisor, line 16: Some use of `echo` could be useful to inform the user of what steps are happening / etc

scripts/post-upgrade.sh
```diff
@@ -0,0 +1,46 @@
...
+kubectl describe PravegaCluster $name -n $namespace
+currentReplicasPc=`kubectl get PravegaCluster $name -n $namespace -o jsonpath='{.status.currentReplicas}'`
+readyReplicasPc=`kubectl get PravegaCluster $name -n $namespace -o jsonpath='{.status.readyReplicas}'`
+if [ $currentReplicasPc != $readyReplicasPc ]; then
+  exit 1
```
bourgeoisor, line 21: When exiting with non-zero, it could be useful to warn the user of _why_ (in this case, not all the replicas are ready), and _how_ they can fix it / what they need to do to proceed

doc/operator-upgrade.md
SrishT: upgrade of zookeeper operator is not required.. and its already mentioned that installation of bookkeeper operator is a prerequite to the upgrade

scripts/post-upgrade.sh
SrishT: done

doc/operator-upgrade.md
SrishT: done.
Also added default values for the fields `namespace` and `zookeeper-svc-name`

doc/operator-upgrade.md
SrishT: done

scripts/post-upgrade.sh
SrishT: Added the reason for failure in `echo` commands

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -601,12 +601,19 @@ func (p *PravegaCluster) migrateBookkeeper(srcObj *v1alpha1.PravegaCluster) erro
...
-	err = migrateConfigMap(srcObj, b)
+	configmap := fmt.Sprintf("%s-configmap", srcObj.Name)
+	err = migrateConfigMap(srcObj, b, configmap)
 	if err != nil {
-		log.Fatalf("Error releasing BK CM %v", err)
+		log.Fatalf("Error releasing BK CM %s %v", configmap, err)
 		return err
 	}
-	log.Print("Migrated Bookkeeper ConfigMap.")
+	configmap = nameForBookie(srcObj.Name)
```
anishakj, line 610: why are we migrating config map 2 times here

SrishT: They are 2 different configmaps

anishakj (approved):
LGTM

PR #443, opened 2020-08-19
Issue 435: Fixing Intermittent E2E test failure

### LINKED ISSUE #435: E2E test failing intermittently
**Problem description**
E2E test failing intermittently

**Problem location**
test/e2e folder

**Suggestions for an improvement**
change's needed in basic test and upgrade test

Prabhaker24:
### Change log description
Fixing Intermittent E2E test failure

### Purpose of the change
Fixes #435 

### What the code does
This code creates new tier2, bk cluster, zk cluster  for each e2e test to avoid the intermittent failure caused due to uncleared data

### How to verify it
Run e2e test multiple times and it should not fail even once without a valid reason.


anishakj (changes requested):
Can we pass clustertype (bookkeeper, zookeeper) to GetCluster, WaitforClusterBecomeReady and Delete cluster, so that same function can be resused

test/e2e/resources/bk-config-map.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
   name: bookkeeper-configmap
 data:
   # Configuration values can be set as key-value properties
-  PRAVEGA_CLUSTER_NAME: pravega
+  PRAVEGA_CLUSTER_NAME: pravega-cluster
```
anishakj, line 7: why is this change required

anishakj (approved):
LGTM

PR #444, opened 2020-08-25
Modifying travis to add retry logic

anishakj:

### Change log description

In travis, we were waiting for 5 min for cluster creation to complete. But in some cases cluster creation takes more time.Add a retry logic, and check whether the cluster state is active.

### Purpose of the change

Add retry mechanism to check cluster is active.

### What the code does
Add retry mechanism to check cluster is active.

### How to verify it

Make sure cluster creation is happening.


Prabhaker24 (approved):
LGTM

PR #446, opened 2020-08-28
Issue 445: Reuse values for helm releases in post-upgrade script

### LINKED ISSUE #445: Post-upgrade script does not reuse the existing values for helm releases
### Description
After upgrading the pravega operator from 0.4.5 to 0.5.1 via helm, the post-upgrade script which is used to upgrade the corresponding pravega release does not reuse the values of  its parameters from the existing setup. Instead the defaults of all fields are reset to the defaults that are specified in the helm charts.

### Importance
must have

### Location
scripts/post-upgrade.sh

The `reuse-values` option can be used instead of setting values for each parameter of the helm release.

SrishT:

### Change log description
After upgrading the pravega operator from 0.4.5 to 0.5.1 via helm, the post-upgrade script which is used to upgrade the corresponding pravega release, does not reuse the values of its parameters from the existing setup. Instead, all these parameters are reset to their default values that have been specified in the new helm charts, which is undesirable.

### Purpose of the change
Fixes #445 

### What the code does
- Accepts the bookkeeper replica count as an additional parameter in the post-upgrade script.
- Uses the `reuse-values` option while upgrading the pravega helm chart in order to preserve the properties that were set at the time of initial installation.

### How to verify it
After the operator upgrade completes, run the post-upgrade script in the following way
```
$ ./post-upgrade.sh <name of pravega cluster> <pravega release name> <bookkeeper release name> <version of pravega cluster> <namespace> <zookeeper svc name> <bookkeeper replica count>
```
After this script executes successfully, the pravega release and the newly created bookkeeper release will be created, while retaining their previous configuration.

### DIFF
--- a/doc/operator-upgrade.md
+++ b/doc/operator-upgrade.md
@@ -121,7 +121,7 @@ kubectl describe PravegaCluster
 ```
 Execute the script `post-upgrade.sh` inside the [scripts](https://github.com/pravega/pravega-operator/blob/master/scripts) folder. The format of the command is
 ```
-./post-upgrade.sh <PravegaCluster resource name> <PravegaCluster release name> <BookkeeperCluster release name> <version> <namespace> <zookeeper svc name>
+./post-upgrade.sh <PravegaCluster resource name> <PravegaCluster release name> <BookkeeperCluster release name> <version> <namespace> <zookeeper svc name> <bookkeeper replica count>
 ```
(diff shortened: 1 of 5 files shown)


anishakj (approved):
LGTM

Prabhaker24 (approved):
LGTM

PR #448, opened 2020-08-31
Issue 447: OpenApiSchemaValidation added

### LINKED ISSUE #447: Need to add OpenApiSchemaValidation
### Description
Need to add OpenApiSchemaValidation

### Importance
should_have

### Location
changes should be added to the crd file in the deploy and charts template folder

add schema validation using kubebuilder controller-gen command

Prabhaker24:

### Change log description
Added Support for OpenApiSchemaValidation 

### Purpose of the change
Fixes #447 

### What the code does
This code adds openApiSchema validation which would check the type of fields and in case of replicas the count also should be above 0 or more

### How to verify it
Try and give a string for replicas field and it should give an error telling replicas is an integer field and can't have string values


anishakj (changes requested):
Can we add more data validation for replicas, pull policy etc as we did in zookeeper-operator

anishakj:
Do we need to generate deepcopy file since crd is changed?

Prabhaker24:
> Do we need to generate deepcopy file since crd is changed?

No we don't need to do that only when we have spec changes i.e we add a new field or remove one existing one we should generate deepcopy.go file.

anishakj:
Please add the additional printer columns back so that kubectl get pravegacluster will displays that

anishakj (approved):
LGTM

SrishT (approved):
LGTM

PR #455, opened 2020-09-29
Issue 452: Correcting setup script and updating documentation

### LINKED ISSUE #452: Pravega setup script does not handle bookkeeper operator certificate creation
### Description
With the upgrade of the bookkeeper operator controller runtime (ref. https://github.com/pravega/bookkeeper-operator/pull/72), the setup script which installs pravega operator and all its dependencies should handle certificate creation for the bookkeeper operator.

### Importance
must have

### Location
setup/pravegacluster.sh

The documentation and commands in the script should also be updated to make use of the published charts for the various pravega dependencies.

SrishT:

### Change log description
To update the setup script to make use of the published charts which installs pravega along with its various dependencies. The failure policy of the webhooks also need to be modified so that the does not allow any pravega deployments if an incorrect certificate has been provided

### Purpose of the change
Fixes #452 

### What the code does
Updates setup script touse published charts. Also modifies the failure policy of the Validating Webhook to make it fail in case an invalid certificate has been provided to it.

### How to verify it
Pravega with all its dependencies should get deployed by providing the following command
```
./pravegacluster.sh deploy
```
Also, if the correct certificate has not been configured while installing the pravega operator, all pravega cluster deployments irrespective of their version should fail with the following error
```
Error: Internal error occurred: failed calling webhook "pravegawebhook.pravega.io": Post https://pravega-webhook-svc.default.svc:443/validate-pravega-pravega-io-v1beta1-pravegacluster?timeout=30s: x509: certificate signed by unknown authority
```


anishakj (approved):
LGTM

PR #456, opened 2020-09-30
Issue-316-added support for DomainName changes

### LINKED ISSUE #316: Pravega Spec parameters nor updatable at runtime
### Description

A lot of Pravega Spec parameters cannot be updated once the cluster is created. This is true of almost all fields other than replicas and version. While not allowing updates on some parameters makes sense, it is necessary to allow updates in a large number of cases like bookkeeper-options, pravega-options, external DNS domain name etc...

### Importance
must-have

Operator code needs to be changed so that updates to necessary fields are allowed during the lifetime of the cluster.

Prabhaker24:

### Change log description
This pr allows the changes in DomainName after the cluster is deployed and changes the **publishedIPAddress** for the pod post the change in the **domainName** filed in the cluster. 

### Purpose of the change
Fixes #316 #463 

### What the code does
This Pr allows the user to change the domainName once the cluster is installed and will change the publishedIPAddress of the pod after that.

### How to verify it
1) Install a pravega cluster without the domain name and update it. It should change the publishedIPAddress  for the ss pods post that
2) Install a pravegacluster with some domainName and update the domainName  It should change the published IP address  for the ss pods post that.
I have checked both the scenarios.


anishakj (approved):
LGTM

PR #462, opened 2020-10-12
issue-464:Added support for JvmOption customization

### LINKED ISSUE #316: Pravega Spec parameters nor updatable at runtime

### LINKED ISSUE #464: Pravega Spec parameters for options and jvm options are not updatable at runtime
### Description
Currently, changes to JVMOptions and options do not take effect during runtime we need to support those for the user to be able to change these fields on runtime.

### Importance
Must_Have

### Location
Pravegacluster_controller.go

Following improvements are needed
1) Restructure the reconcile logic and add reconcile for configmap
2) Allow upgrade for config map for any change in either JVMOption or Option
3) Change the validation webhook to not allow any invalid change to option or JVMOption to go to reconcile

Prabhaker24:

### Change log description
This pr will enable the user to modify both controller and sgmentstore JVMOptions as well as options during runtime

### Purpose of the change
Fixes #464 

### What the code does
The code looks for any change in the jvm option or option in either of controller or segmentstore and if there is a change i.e user has changed any of the jvmOptions or options or has added a new value, in that case, we update the configmap and restart each pod one by one.
I am doing validation as well and is done by the method **validateConfigMap()** which ensures user doesn't change any of the following values:-
1) controller.containerCount
2) controller.container.count
3) pravegaservice.containerCount
4) pravegaservice.container.count
5) bookkeeper.bkLedgerPath
6) bookkeeper.ledger.path
7) controller.retention.bucketCount
8) controller.retention.bucket.count
9) controller.watermarking.bucketCount
10) controller.watermarking.bucket.count
11) pravegaservice.dataLogImplementation
12) pravegaservice.dataLog.impl.name
13) pravegaservice.storageImplementation
14) pravegaservice.storage.impl.name
15) storageextra.storageNoOpMode
16) storageextra.noOp.mode.enable

### How to verify it
1) Deploy Pravega cluster and update segmentstore JVM Option we should see that segmentstore configmap should get updated and each pod should restart sequentially.
2)  Deploy Pravega cluster and update controller JVM Option we should see that controller configmap should get updated and each pod should restart sequentially.
3) Deploy Pravega cluster and update Option we should see that controller and segmentstore configmap should get updated and each of the controller and ss pod should restart sequentially.
4) Try changing any of the above 16 values that the user is not supposed to change and the operator should error saying that the parameter is not supposed to be changed.

### DIFF
--- a/pkg/controller/pravegacluster/pravegacluster_controller.go
+++ b/pkg/controller/pravegacluster/pravegacluster_controller.go
@@ -369,7 +369,7 @@ func (r *ReconcilePravegaCluster) deployController(p *pravegav1beta1.PravegaClus
 		if err != nil {
 			return err
 		}
-		err = r.restartDeploymentPod(podList)
+		err = r.restartDeploymentPod(podList, p)
 		if err != nil {
 			return err
 		}
@@ -532,7 +532,7 @@ func (r *ReconcilePravegaCluster) restartStsPod(podList *corev1.PodList) error {
 	}
 	return nil
 }
-func (r *ReconcilePravegaCluster) restartDeploymentPod(podList *corev1.PodList) error {
+func (r *ReconcilePravegaCluster) restartDeploymentPod(podList *corev1.PodList, p *pravegav1beta1.PravegaCluster) error {
 	for _, podItem := range podList.Items {
 		err := r.client.Delete(context.TODO(), &podItem)
 		if err != nil {
@@ -546,6 +546,22 @@ func (r *ReconcilePravegaCluster) restartDeploymentPod(podList *corev1.PodList) 
 					return fmt.Errorf("failed to delete controller pod (%s) for 10 mins ", podItem.ObjectMeta.Name)
 				}
 				err = r.client.Get(context.TODO(), types.NamespacedName{Name: podItem.ObjectMeta.Name, Namespace: podItem.ObjectMeta.Namespace}, pod)
+			}
+			deploy := &appsv1.Deployment{}
+			name := p.DeploymentNameForController()
+			err = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, deploy)
+			if err != nil {
+				return fmt.Errorf("failed to get deployment (%s): %v", deploy.Name, err)
+			}
+			start = time.Now()
+			for deploy.Status.ReadyReplicas != deploy.Status.Replicas {
+				if time.Since(start) > 10*time.Minute {
+					return fmt.Errorf("failed to make controller pod ready for 10 mins ")
+				}
+				err = r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: p.Namespace}, deploy)
+				if err != nil {
+					return fmt.Errorf("failed to get deployment (%s): %v", deploy.Name, err)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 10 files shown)


pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -316,11 +317,32 @@ func (r *ReconcilePravegaCluster) deployController(p *pravegav1beta1.PravegaClus
...
 	}
 
+	var eq bool = true
+	currentConfigMap := &corev1.ConfigMap{}
 	configMap := pravega.MakeControllerConfigMap(p)
 	controllerutil.SetControllerReference(p, configMap, r.scheme)
-	err = r.client.Create(context.TODO(), configMap)
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return err
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: configMap.Name, Namespace: p.Namespace}, currentConfigMap)
```
pbelgundi, line 324: This does not seem like a good place and a good way to check for updates to config map.
Please follow the logic used in zookeeper -operator:
https://github.com/pravega/zookeeper-operator/blob/master/pkg/controller/zookeepercluster/zookeepercluster_controller.go#L150

In p-operator we're checking very specific things in the reconcile loop, instead of checking for updates to each component and that is complicating the flow of code way too much...
Lets refactor and change it to the way it is done in ZK operator...

pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -329,6 +351,29 @@ func (r *ReconcilePravegaCluster) deployController(p *pravegav1beta1.PravegaClus
 	if err != nil && !errors.IsAlreadyExists(err) {
 		return err
 	}
+	if !eq {
```
pbelgundi, line 354: Same comment adding too many ifs... complicates the flow of code.. I think its time to refactor the reconcile loop...

pbelgundi (changes requested):

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: Done.

pkg/controller/pravegacluster/pravegacluster_controller.go
Prabhaker24: Done.

pbelgundi (approved):
LGTM. Just make sure we add enough tests to cover this.

test/e2e/cmchanges_test.go
```diff
@@ -0,0 +1,93 @@
...
+	g.Expect(err).NotTo(HaveOccurred())
+
+	//updating pravega option
+	pravega.Spec.Pravega.Options["bookkeeper.bkAckQuorumSize"] = "2"
+
+	//updating pravegacluster
+	err = pravega_e2eutil.UpdatePravegaCluster(t, f, ctx, pravega)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	//checking if the upgrade of options was successfull
```
anishakj, line 66: Fix the typo `successful`

Prabhaker24: Done.

anishakj (approved):
LGTM

PR #469, opened 2020-10-20
Issue-468: Run pravega segmentstore sts with runAsUser = 0

### LINKED ISSUE #468: Run pravega segmentstore sts with runAsUser: 0
### Description

Pravega with ECS as LTS is failing to deploy in Openshift environment, as container needs to inject cert bundle into trust store, but the security context selected by openshift is preventing it.

### Importance
must_have

### Location
pravega_segmentstore.go

somewhere around https://github.com/pravega/pravega-operator/blob/569dfb38a6cb35d246aa2b32ffdd60e56e4f8502/pkg/controller/pravega/pravega_segmentstore.go#L107 add the following.
``` spec:
  securityContext:
  runAsUser: 0 ```

### PR DESCRIPTION
**Pull request by Prabhaker24**

### Change log description
Pravega with ECS as LTS is failing to deploy in Openshift environment, as container needs to inject cert bundle into trust store, but the security context selected by openshift is preventing it.
To solve the above problem we have to set runAsUser = 0 for  pravega segmentstore sts.

### Purpose of the change
Fixes #468 

### What the code does
These changes allow the segmentstore pod's to run as root user. For enabling this user has to give the following things in the manifest file while deploying it manually

```
 pravega:
   segmenStoreSecurityContext: 
    runAsUser: 0  
```
If he is using charts he needs to uncomment the line in the values.yaml in the pravega charts:-

```
segmentStoreSecurityContext: 
  runAsUser: 0

```
For controller he needs to set like this:-

```
controllerSecurityContext:
   runAsUser: 0
```

### How to verify it
Deploy the pravega cluster with **runAsUser: 0** and check the segment store sts and controller deployment post-deployment  it should have the following in it's described output

```
      securityContext:
        runAsUser: 0
```

### DIFF
--- a/charts/pravega/README.md
+++ b/charts/pravega/README.md
@@ -99,4 +99,5 @@ The following table lists the configurable parameters of the pravega chart and t
 | `storage.cache.className` | Storage class for cache volume | `` |
 | `storage.cache.size` | Storage requests for cache volume | `20Gi` |
 | `options` | List of Pravega options | |
-| `securityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` | 
+| `segmentStoreSecurityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` |
+| `controllerSecurityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` |
--- a/charts/pravega/README.md
+++ b/charts/pravega/README.md
@@ -99,3 +99,4 @@ The following table lists the configurable parameters of the pravega chart and t
 | `storage.cache.className` | Storage class for cache volume | `` |
 | `storage.cache.size` | Storage requests for cache volume | `20Gi` |
 | `options` | List of Pravega options | |
+| `securityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` | 
--- a/charts/pravega/README.md
+++ b/charts/pravega/README.md
@@ -99,5 +99,5 @@ The following table lists the configurable parameters of the pravega chart and t
 | `storage.cache.className` | Storage class for cache volume | `` |
 | `storage.cache.size` | Storage requests for cache volume | `20Gi` |
 | `options` | List of Pravega options | |
-| `segmentStoreSecurityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` |
-| `controllerSecurityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` |
+| `segmentStoreSecurityContext` | Holds pod-level security attributes and common container settings | `{}` |
+| `controllerSecurityContext` | Holds pod-level security attributes and common container settings | `{}` |
--- a/deploy/crds/pravega.pravega.io_pravegaclusters_crd.yaml
+++ b/deploy/crds/pravega.pravega.io_pravegaclusters_crd.yaml
@@ -290,110 +290,9 @@ spec:
                           to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                         type: object
                     type: object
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

File: pkg/apis/pravega/v1beta1/pravega.go [hunk 1]
```diff
@@ -167,6 +168,9 @@ type PravegaSpec struct {
 
 	// SegmentStoreExternalTrafficPolicy defines the ExternalTrafficPolicy it can have cluster or local
 	SegmentStoreExternalTrafficPolicy string `json:"segmentStoreExternalTrafficPolicy,omitempty"`
+
+	// SecurityContext holds security configuration that will be applied to a container
+	SecurityContext *corev1.PodSecurityContext `json:"securityContext,omitempty"`
```
sarlaccpit, line 173: The PodSecurityContext is getting applied to the segment store only in this PR.     And indeed the controller pods don't necessarily need the same PSC.    (in particular, SegmentStores pod have more requirements in terms of requiring access to NFS for example, versus controllers which require less).

While the changes look correct, perhaps calling out that this is only getting applied to the SS, (renaming it accordingly?) , or also adding a separate one for the controller pods as well, might help avoid confusion?

test/e2e/resources/crd.yaml
```diff
@@ -392,6 +392,128 @@ spec:
                       to the Pravega processes as JAVA_OPTS. See the following file
                       for a complete list of options: https://github.com/pravega/pravega/blob/master/config/config.properties'
                     type: object
+                  securityContext:
```
gaddas3, line 395: Is this from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#securitycontext-v1-core?

deploy/crds/pravega.pravega.io_pravegaclusters_crd.yaml
```diff
@@ -392,6 +392,128 @@ spec:
                       to the Pravega processes as JAVA_OPTS. See the following file
                       for a complete list of options: https://github.com/pravega/pravega/blob/master/config/config.properties'
                     type: object
+                  securityContext:
```
gaddas3, line 395: Is this from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#securitycontext-v1-core?

gaddas3 (approved):

deploy/crds/pravega.pravega.io_pravegaclusters_crd.yaml
Prabhaker24: yes, it's from this only.

pkg/apis/pravega/v1beta1/pravega.go
Prabhaker24: I have changed the name and added securitycontext for controller as well.

charts/pravega/README.md
```diff
@@ -99,3 +99,5 @@ The following table lists the configurable parameters of the pravega chart and t
 | `storage.cache.className` | Storage class for cache volume | `` |
 | `storage.cache.size` | Storage requests for cache volume | `20Gi` |
 | `options` | List of Pravega options | |
+| `segmentStoreSecurityContext.runAsUser` | The UID to run the entrypoint of the container process | `0` |
```
anishakj, line 102: Can we mention that by default we are not setting any value for security context.

anishakj (changes requested):

charts/pravega/README.md
Prabhaker24: Done.

anishakj (approved):
LGTM

gaddas3 (approved):

sarlaccpit (approved):

Issue #470, opened 2020-10-28
Inconsistencies with the webhook cert name can lead to issues accessing it


sarlaccpit:
### Description

Inconsitency between the two paths where the webhook cert is generated or provided can cause SSL isues when accessing the webhook:

```
 Internal error occurred: failed calling webhook "pravegawebhook.pravega.io": Post https://pravega-webhook-svc.nautilus-pravega.svc:443/validate-pravega-pravega-io-v1bet
                                          a1-pravegacluster?timeout=30s: x509: certificate signed by unknown authority
```

The problem is that in the case where the cert needs to be generated ( `.Values.webhookCert.generate: true`), the `certName` field https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator/values.yaml#L37 needs to be set (despite the comment above it) and match what is used here https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator/templates/certificate.yaml#L18, else this does not work https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator/templates/webhook.yaml#L26

The reason being line 26 does not know if the cert was generated or not.  It just needs the name.

### Importance

There is a workaround, which is to ignore the comment in the value file and set the certName to match https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator/templates/certificate.yaml#L18

Therefore this is not a blocking issue but can be misleading to others who may run into the same problem.

### Location

This is a charts issue only.  See links in the description for specific locations.

### Suggestions for an improvement

The solution is to add a template helper to hide this nuance so that at the end of the day, whether the cert was generated or provided, the name is correct for what the webhookconfiguration consumes here https://github.com/pravega/pravega-operator/blob/master/charts/pravega-operator/templates/webhook.yaml#L26

anishakj:
Fixed by PR https://github.com/pravega/pravega-operator/pull/475

PR #475, opened 2020-11-05
Issue 473: Fixing charts to set the required LTS type

### LINKED ISSUE #473: Unable to set ECS or HDFS as tier2 using the latest charts
### Description
With the latest charts, pravega is not being deployed with either ECD or HDFS as tier2 and it gets defaulted to filesystem in all cases.

### Importance
must have

### Location
helm charts

Fix the charts to take correct configuration based on the longtermstorage type that has been set by the user in the values.yaml

### LINKED ISSUE #474: Helm upgrade is failing to update fields  after an upgrade failure
### Description

After installing pravega, tries to update an option field whose update is not allowed 
```
helm upgrade pravega charts/pravega --set-string options."pravegaservice\.container\.count"="10"
Error: UPGRADE FAILED: cannot patch "pravega" with kind PravegaCluster: admission webhook "pravegawebhook.pravega.io" denied the request: controller.container.count should not be changed 
```
After that tries to update any other filed, gives the same error message and it is not getting updated
```
 helm upgrade pravega charts/pravega --set-string options."log\.level"="DEBUG"
Error: UPGRADE FAILED: cannot patch "pravega" with kind PravegaCluster: admission webhook "pravegawebhook.pravega.io" denied the request: controller.container.count should not be changed 
```

### Importance

should-have

### Location

Charts

Fix charts

SrishT:

### Change log description
Currently even when the longTermStorageType is set to `ecs` or `hdfs` within the values.yaml, the ecs or hdfs configuration is not taking effect and the longTermStorageType  of the deployed pravega cluster is getting set to filesystem in all cases.

### Purpose of the change
Fixes #473  , #470 

### How to verify it
Following use-cases have been tested :
- Deployed pravega cluster with storageType = ecs
- Deployed pravega cluster with storageType = hdfs
- Deployed pravega operator 0.4.5 with v1alpha1 pravega cluster. Upgraded the pravega operator to 0.5.2 followed by upgrading to v1beta1 pravega cluster.
- Deployed pravega operator by setting webhookCert.generate within values.yaml to true and confirmed that the correct certificate name is used.

### DIFF
--- a/charts/pravega-operator/Chart.yaml
+++ b/charts/pravega-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: pravega-operator
 description: Pravega Operator Helm chart for Kubernetes
-version: 0.5.1
-appVersion: 0.5.1
+version: 0.5.2
+appVersion: 0.5.2
 keywords:
 - pravega
 - storage
--- a/charts/pravega-operator/README.md
+++ b/charts/pravega-operator/README.md
@@ -51,7 +51,7 @@ The following table lists the configurable parameters of the pravega-operator ch
 | Parameter | Description | Default |
 | ----- | ----------- | ------ |
 | `image.repository` | Image repository | `pravega/pravega-operator` |
-| `image.tag` | Image tag | `0.5.1` |
+| `image.tag` | Image tag | `0.5.2` |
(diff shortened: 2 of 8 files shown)


anishakj (approved):
LGTM

PR #478, opened 2020-11-11
Issue 477: Handling Segmentstore STS upgrade

### LINKED ISSUE #477: Pravega Operator post controller-runtime upgrade does not handle Segment Store STS upgrade one at a time
### Description
Pravega Operator adopts an `OnDelete` upgrade strategy to handle segmentstore upgrade in which segmentstore pods are brought down one at a time and the updated pod comes back up. However from Pravega Operator version 0.5.0 onwards, it has been observed that multiple segmentstore pods are being brought down when an upgrade is triggered, which might be detrimental.

### Importance
must have

### Location
pkg/controller/pravegacluster/upgrade.go

SrishT:

### Change log description
With the upgrade of controller runtime, the updated `client.List` method does not return the list of pods in a sorted order, which was breaking the sequential one-pod-at-a-time upgrade logic that had been implemented in the operator for the segmentstore statefulset.

### Purpose of the change
Fixes #477 

### What the code does
The list of pods returned by the `client.List` method is sorted by name before doing any further processing, to maintain consistency with the previous segmentstore upgrade logic.

### How to verify it
Upgrade the pravega cluster from one version to another and ensure that not more than one segmentstore pod is upgraded at any given time.

### DIFF
--- a/pkg/controller/pravegacluster/upgrade.go
+++ b/pkg/controller/pravegacluster/upgrade.go
@@ -13,6 +13,7 @@ package pravegacluster
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
 	pravegav1beta1 "github.com/pravega/pravega-operator/pkg/apis/pravega/v1beta1"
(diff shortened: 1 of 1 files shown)


anishakj (approved):
LGTM

PR #492, opened 2021-01-03
Issue 491: Handling config key name changes

### LINKED ISSUE #491: Pravega webhook fails upgrade even when the name of the container count property is modified as per the new naming convention
### Description
Pravega webhook fails upgrade even when just the name of the container count property is modified as per the new naming convention.

### Importance
must have

### Location
pkg/apis/pravega/v1beta1/pravegacluster_types.go

While the value of this property should never be allowed to change, the name of this property needs to change when upgrading pravega version to above 0.9.0

SrishT:

### Change log description
Pravega webhook fails upgrade even when just the name of the configuration keys is modified as per the new naming convention (ref. https://github.com/pravega/pravega/pull/4713)

### Purpose of the change
Fixes #491 

### How to verify it
Upgrade pravega from version 0.7.0 to 0.8.0 and along with it also modify the name of the pravega option name from `controller.containerCount` to `controller.container.count` without modifying its value. The upgrade should be allowed and you should not see the following error message
```
cannot patch "nautilus" with kind PravegaCluster: admission webhook "pravegawebhook.pravega.io" denied the request: controller.container.count should not be changed
```

### DIFF
--- a/pkg/apis/pravega/v1beta1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1beta1/pravegacluster_types.go
@@ -1028,132 +1028,205 @@ func (p *PravegaCluster) validateConfigMap() error {
 			return fmt.Errorf("failed to get configmap (%s): %v", configmap.Name, err)
 		}
 	}
+	data := strings.Split(configmap.Data["JAVA_OPTS"]," ")
+	eq := false
 	if val, ok := p.Spec.Pravega.Options["controller.containerCount"]; ok {
-		checkstring := fmt.Sprintf("-Dcontroller.containerCount=%v", val)
-		eq := strings.Contains(configmap.Data["JAVA_OPTS"], checkstring)
+		key := fmt.Sprintf("-Dcontroller.containerCount=%v", val)
+		for _, checkstring := range data {
+			if checkstring == key {
+				eq = true
+			}
+		}
 		if !eq {
-			return fmt.Errorf("controller.containerCount should not be changed ")
+			return fmt.Errorf("controller.containerCount should not be modified")
 		}
 	}
+	eq = false
 	if val, ok := p.Spec.Pravega.Options["controller.container.count"]; ok {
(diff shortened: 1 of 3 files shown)


pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1065,8 +1069,10 @@ func (p *PravegaCluster) validateConfigMap() error {
 	}
 	if val, ok := p.Spec.Pravega.Options["bookkeeper.ledger.path"]; ok {
 		checkstring := fmt.Sprintf("-Dbookkeeper.ledger.path=%v", val)
-		eq := strings.Contains(configmap.Data["JAVA_OPTS"], checkstring)
-		if !eq {
+		eq_new := strings.Contains(configmap.Data["JAVA_OPTS"], checkstring)
```
Prabhaker24, line 1072: In place of contains let's check if the configmap.Data["JAVA_OPTS"] is equal to the checkstring or not to avoid any false positives.
Please change this at all other places too.

Prabhaker24 (changes requested):

pkg/apis/pravega/v1beta1/pravegacluster_types.go
SrishT: We cannot use the equals check here since this particular option is just one of the options within the list of java options for pravega.

Prabhaker24: what I meant is let's think of a logic where we should check if specific option field value is equal to the changed value or not.
for eg think of a case where a user tries to change ledger path from **ledgers/temp1 to ledgers**   this will pass our check even though the ledger path is changed.

Prabhaker24 (changes requested):

Prabhaker24 (approved):
LGTM

anishakj (approved):
LGTM

PR #494, opened 2021-01-07
Issue 493: Providing upgrade path to a test version

### LINKED ISSUE #493: Add support to provide a custom upgrade path to a test version
### Description
Currently the pravega operator provides support to install a test version which has not yet been released and is therefore not a part of the supported versions config map. However, there is no way today to supply a custom upgrade path in the supported versions configmap, which could allow testing out any upgrade paths including to non released versions.

### Importance
This issue is not a blocker as there is a workaround to achieve this by editing the supported-versions configmap. However it is something that we should definitely add support for.

### Location
charts

We could add two values under test mode,  `fromVersion` and `toVersion`.

SrishT:

### Change log description
This PR adds support to supply a custom upgrade path, which would enable testing the installation and upgrade of the pravega cluster to non-released test versions.

### Purpose of the change
Fixes #493 

### What the code does
When the user enables `testmode` and provides both the `fromVersion` and `toVersion` inside the values.yaml, it does either of the following :
- if an entry corresponding to the fromVersion is already present in the supported-versions configmap, the entry corresponding to the fromVersion in the supported-versions configmap is replaced with a fresh entry which provides an upgrade path from fromVersion to toVersion
- if no entry corresponding to the fromVersion exists in the supported-versions configmap, a new entry is created which provides an upgrade path from fromVersion to toVersion

Also when the user enables testmode and specifies just the fromVersion, a new entry is created in the supported-versions configmap, which allows this version to be installed

### How to verify it
- Inside the values.yaml, set the value of testmode.enabled to true and provide values for testmode.fromVersion and testmode.toVersion. You should be able to see an entry of the following format added to the supported-version configmap
```
<fromVersion> : <fromVersion> , <toVersion>
```
- set the value of testmode.enabled to true and provide a value only for testmode.fromVersion and leave testmode.toVersion as is. You should be able to see an entry of the following format added to the supported-version configmap
```
<fromVersion> : <fromVersion>
```
- set the value of testmode.enabled to false. In this case you should not see any changes in the contents of the supported-version configmap


anishakj (changes requested):
Can we add `0.7.3` as well as it is published in dockerhub

jdmaguire:
The number of files/arguments/conditional branches could be reduced if the values took something like this:

```
versionMap:
  "0.1.0" : ["0.1.0"]
  "0.2.0" : ["0.2.0"]
  "0.3.0" : ["0.3.0", "0.3.1", "0.3.2"]
  "0.3.1" : ["0.3.1", "0.3.2"]
......
```

Thoughts?

The code in this PR seems pretty complicated for a basic "I want to upgrade to a custom version tag."

jkhalack:
I agree with @jdmaguire 
Having a custom upgrade path supplied by user would be much more clear
> versionMap:
>    "0.8.0": ["0.7.0", "0.8.0"]

SrishT:
@jdmaguire providing the upgrade path in the way you have mentioned would require a change in the operator code, and would still not change the conditional checks done here since we need to replace an entry if the version mentioned in `fromVersion` is already a part of the version map.

jkhalack (approved):

anishakj (approved):
LGTM

PR #501, opened 2021-01-22
Issue 500: Adding support for test versions

### LINKED ISSUE #500: Add support to install as well as provide upgrade paths to unreleased test versions
### Description
Enabling testmode currently wither allows one to express an upgrade to an unreleased version OR a allows straight installation of an unreleased version, but not both. We need a way that allows a single value file to express both possibilities.

### Importance
must have

### Location
charts

the value specified in `testmode.version` should be added to the version-map to allow deployment of this version as well as to allow upgrade from any version to this version. Furthermore, the `testmode.fromVersion` can be persisted to allow upgrade from any unreleased version to the unreleased version `testmode.version`

SrishT:

### Change log description
This PR adds support to provide an upgrade path as well as to add an entry corresponding to an unreleased test version, which would enable testing both installation and upgrade of the pravega cluster to non-released test versions.

### Purpose of the change
Fixes #500

### What the code does
When the user enables `testmode` and provides `version` within the values.yaml
- the version is appended to the upgrade path of each version within the version map, thereby allowing an upgrade from any version to this version.
- a new entry is created in the version map corresponding to this version, which allows this version to be installed.

When additionally the user also provides a value for `fromVersion` in addition to the above values
- a new entry is created to provide an upgrade path from `fromVersion` to `version`

### How to verify it
- Inside the values.yaml, set the value of testmode.enabled to true and provide values for `testmode.version`. This should append `testmode.version` to each of the entries in the supported-versions configmap, and should also create a new entry corresponding to this version
```
<key 1> : <value 1.1> , ... , <version>
<key 2> : <value 2.1> , ... , <version>
....
....
<version>: <version>
```
- Now also provide a value for `testmode.fromVersion` in addition to the above values. This should create an entry of the following format added to the supported-versions configmap
```
<fromVersion> : <fromVersion> , <version>
```
- set the value of testmode.enabled to false. In this case you should not see any changes in the contents of the supported-versions configmap


jkhalack (approved):
LGTM

anishakj (approved):
LGTM

PR #503, opened 2021-02-05
Issue 502: Pravega Operator pre-delete hook

### LINKED ISSUE #502: No pre-delete hook in pravega operator chart
### Description
The pravega operator helm chart does not have a pre-delete hook in it. As a result, the operator can be uninstalled with orphan custom resources hanging around.
A pre-delete hook needs to be added to the operator chart that would verify that all custom resources have been cleaned up before deleting the operator.

### Importance
should have

### Location
pravega operator helm charts

SrishT:
### Change log description
Provides a pre-delete hook which ensures that custom resources are deleted before deleting the operator.

### Purpose of the change
Fixes #502 

### What the code does
Allows the operator to be deleted via helm only if the PravegaCluster custom resource has already been deleted, not otherwise.

### How to verify it
- If you try to delete the pravega operator before having deleted the PravegaCluster custom resource, the pre-delete job will fail, thereby not allowing the operator from being deleted.
- If you try to delete the pravega operator after the PravegaCluster custom resource has been cleaned up successfully, the operator will also be successfully deleted.

### DIFF
--- a/charts/pravega-operator/templates/pre-delete-hooks.yaml
+++ b/charts/pravega-operator/templates/pre-delete-hooks.yaml
@@ -1,0 +1,110 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: {{ template "pravega-operator.fullname" . }}-pre-delete
+  annotations:
+    "helm.sh/hook": pre-delete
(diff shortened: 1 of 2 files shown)


anishakj (approved):
LGTM

PR #506, opened 2021-02-10
Issue-505: Added changes for Avoiding 2 Restarts when Both Version and cm changes are done at the same time 

### LINKED ISSUE #505: Error message is seen while upgrading Pravega Cluster along with CM Changes in Pravega cluster
### Description

Error messages are seen while upgrading the Pravega cluster along with CM Changes in Pravega cluster due to Rae condition

### Importance
nice_to_have

### Location
pravegacluster_controller.go

Try to avoid upgrade of pravega cluster and CM at the same time by skipping restart pod for cm changes in case both upgrades of version and cm both are happening as eventually version upgrade does that.

Prabhaker24:

### Change log description
Added changes for Avoiding 2 Restarts when Both Version and cm changes are done at the same time. This makes the upgrade process fast and we don't get error messages due to race condition.

### Purpose of the change
Fixes #505 

### What the code does

Code checks for any change in the version of the bk cluster when there are changes in cm and if there is change in the version
then we skip restart of pods and only upgrade the cm

### How to verify it
Install pravega cluster then upgrade the cluster along with changing some field in the option we should only see one restart of all pravega(controller and ss) pods and the upgrade should complete.

### DIFF
--- a/pkg/controller/pravegacluster/pravegacluster_controller.go
+++ b/pkg/controller/pravegacluster/pravegacluster_controller.go
@@ -242,9 +242,11 @@ func (r *ReconcilePravegaCluster) reconcileControllerConfigMap(p *pravegav1beta1
 				return err
 			}
 			//restarting controller pods
-			err = r.restartDeploymentPod(p)
-			if err != nil {
-				return err
+			if !r.checkVersionUpgradeTriggered(p) {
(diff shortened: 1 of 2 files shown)


anishakj (approved):
LGTM

Issue #507, opened 2021-02-16
Extend functionality of Test Mode to enable privileged mode for Segment Store containers


leapsky:
### Description

By default, Docker containers run in unprivileged mode. However, to profile the applications, we use tracers (e.g. `perf_events`, `strace`, `ebpf programs`) that requires more sufficient privileges. We need to extend functionality of `Test Mode` to enable `privileged mode` for Segment Store containers which will grant them operating system administrative capabilities

### Importance

This functionally will allow us to use Linux tracing tools without patching Pravega Operator and make profiling more simple. 

### Location

`pkg/controller/pravega/pravega_segmentstore.go`

### Suggestions for an improvement

Enable docker container `privileged mode` in `Test Mode`

anishakj:
Fixed by PR https://github.com/pravega/pravega-operator/pull/508

PR #508, opened 2021-02-16
Extend functionality of Test Mode to enable privileged mode for Segme…

leapsky:
Extend functionality of Test Mode to enable `privileged mode` for Segment Store containers #507

### Change log description

* Added `SecurityContext` definition in Segment Store containers' description

### Purpose of the change

This functionally will allow us to use Linux tracing tools without patching Pravega Operator and make profiling more simple.

### What the code does

Enable privileged mode for Segment Store containers.

### How to verify it

When Pravega deployment is completed, check if Segment Store containers running in privileged mode:
`kubectl get pod pravega-pravega-segment-store-0 -o yaml | grep privileged`

### DIFF
--- a/charts/pravega-operator/templates/pravega.pravega.io_pravegaclusters_crd.yaml
+++ b/charts/pravega-operator/templates/pravega.pravega.io_pravegaclusters_crd.yaml
@@ -956,11 +956,6 @@ spec:
                           specified in SecurityContext takes precedence for that container.
                         format: int64
                         type: integer
-                      privileged:
-                        description: Indicates that the container must run in unprivileged or privileged mode.
-                          If true, the container is "privileged", and it is given access to all devices on the host.
-                          If false, the container is not allowed to access any devices on the host.
-                        type: boolean
                       seLinuxOptions:
                         description: The SELinux context to be applied to all containers.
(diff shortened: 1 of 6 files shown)


anishakj (changes requested):
We are setting security context for the pod at  https://github.com/pravega/pravega-operator/blob/master/pkg/controller/pravega/pravega_segmentstore.go#L169. Can't we set the `privileged` mode there?

leapsky:
We could use, but in `segmentStoreSecurityContext`, there is no `privileged` among the described properties.

https://github.com/pravega/pravega-operator/blob/master/deploy/crds/pravega.pravega.io_pravegaclusters_crd.yaml#L1835

anishakj:
> We could use, but in `segmentStoreSecurityContext`, there is no `privileged` among the described properties.
> 
> https://github.com/pravega/pravega-operator/blob/master/deploy/crds/pravega.pravega.io_pravegaclusters_crd.yaml#L1835

Got it, both of the security context structures are different. Could you fix the formatting errors

leapsky:
Sure, will do.

anishakj (approved):
LGTM

PR #514, opened 2021-03-10
Issue 515: add customized labels support

realAaronWu:
### Change log description

add controllerPodLabels and segmentStorePodLabels field in the crd, each one is a map with key as the label key and value as the label value.

### Purpose of the change

issue 515

### What the code does

add labels to the controller deployment and pods, and segmentstore sts and pods

### How to verify it

deploy pravega with labels set see if labels are attached to pods


anishakj (changes requested):
It looks like unit tests are failing. Could you please fix that?

anishakj (approved):
LGTM

PR #524, opened 2021-04-06
Issue 525: Add suport of configurable extra configmap volume mounts

CraneShiEMC:
### Change log description

- introduces the `configMapVolumeMounts` option as multiple key-value pairs, separated by commas and each consisting of a `<configMapName>`:`<configKey>`=`<configFilePath>` tuple.
- parses the `configMapVolumeMounts` values.
- adds the extracted values to the `volumes` list and containers' `volumeMounts` list.

### Purpose of the change

Implements #525 

Pravega operator should allow to specify "extra" configmap volumes to be mounted into pravega pods, including pravega segment store pods and pravega controller pods. Such config volumes may be used to import external configs for pravega use.

### What the code does

The code implements the new pravega option `configMapVolumeMounts`.
To use "extra" volumes in the pravega pods, in the Pravega options, specify:

- `configMapVolumeMounts` to add `configMap` volumes

The option consist of multiple key-value pairs, separated by commas and each consisting of a `<configMapName>`:`<configKey>`=`<configFilePath>` tuple:

`<configMapName>` is the name of the configmap to be mounted.
`<configKey>` is the key (file name) of the configmap data.
`<configFilePath>` is the path of the config file inside the container to be overriden..
For example, `configMapVolumeMounts: "prvg-logback:logback.xml=/opt/pravega/conf/logback.xml"`

### How to verify it

1. Add the `configMapVolumeMounts` option to Pravega options parameter, e.g., `configMapVolumeMounts: "prvg-logback:logback.xml=/opt/pravega/conf/logback.xml"`.
2. When the pravega cluster is deployed and ready, make sure that the the `volumes` list and containers' `volumeMounts` list has contained the configMapVolumeMounts volumes and mounts specified in the `configMapVolumeMounts`. In addition, make sure that the config file inside the pravega pods has been overriden by the data in the config map.

The function has worked as expected on the deployed pravega cluster in our environment. We've already successfully both the positive and negative test cases. In the negative case, I've also verified successfully the scenario that the pravega is install successfully as originally expected without all the customized pod labels and extra volume mounts parameters.

### DIFF
--- a/pkg/controller/pravega/pravega_controller.go
+++ b/pkg/controller/pravega/pravega_controller.go
@@ -60,19 +60,17 @@ func MakeControllerPodTemplate(p *api.PravegaCluster) corev1.PodTemplateSpec {
 }
 
 func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
+	// Parse volumes & volumeMounts parameters first
 	var hostPathVolumeMounts []string
 	var emptyDirVolumeMounts []string
+	var configMapVolumeMounts []string
 	var ok bool
+
+	var volumes []corev1.Volume
+	var volumeMounts []corev1.VolumeMount
 
 	if _, ok = p.Spec.Pravega.Options["hostPathVolumeMounts"]; ok {
 		hostPathVolumeMounts = strings.Split(p.Spec.Pravega.Options["hostPathVolumeMounts"], ",")
-	}
-	if _, ok = p.Spec.Pravega.Options["emptyDirVolumeMounts"]; ok {
-		emptyDirVolumeMounts = strings.Split(p.Spec.Pravega.Options["emptyDirVolumeMounts"], ",")
-	}
-
-	var volumes []corev1.Volume
-	if len(hostPathVolumeMounts) > 1 {
 		for _, vm := range hostPathVolumeMounts {
 			s := strings.Split(vm, "=")
 			v := corev1.Volume{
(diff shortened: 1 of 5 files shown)


anishakj (changes requested):
@CraneShiEMC Code coverage has gone down, for the changes newly added. Could you please fix those ?

CraneShiEMC:
> @CraneShiEMC Code coverage has gone down, for the changes newly added. Could you please fix those ?

@anishakj Thank you very much for your information. I will work on it.

pkg/controller/pravega/pravega_controller.go
```diff
@@ -106,6 +117,37 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
...
+					},
+				},
+			}
+			volumes = append(volumes, v)
+
+			m := corev1.VolumeMount{
+				Name:      s[0],
+				MountPath: p[1],
+				SubPath:   s[1],
+			}
```
SrishT, line 148: could u please explain why is the subpath being set here?
could we not instead set the volume mount in the following manner
```
Name:      s[1],
MountPath: p[1],
```

CraneShiEMC: Hi Srishti, if we don't use subPath and only use the name and mountPath for the configMap volume mount here, the ConfigMap value, i.e. content of our version of logback conf, will not override the file specified in `MountPath`, i.e. `/opt/pravega/conf/logback.xml`, as expected. Instead, it will override the file path `MountPath\<configKey>`, i.e. `/opt/pravega/conf/logback.xml/logback.xml`, which is not what we expected.

On the other hand, there are many existing pravega confs in `/opt/pravega/conf/` and we only want to override the logback conf `/opt/pravega/conf/logback.xml`.

To meet our requirment, the use of subpath here is our only choice.

anishakj (approved):
LGTM

PR #526, opened 2021-04-07
Issue 527: Support of multiple containers in pravega pods

### LINKED ISSUE #527: Support of multiple containers in pravega pods
### Description

We have a logic to check pod faulty when doing upgrade, but currently the logic only check the first container in the pod, which does not necessarily to be the pravega container, when customer inject sidecars.

### Importance

should-have

### Location

Pravegacluster

Loop the containers and pick the correct container to check faulty.

realAaronWu:
### Change log description

We have a logic to check pod faulty when doing upgrade, but currently the logic only check the first container in the pod, which does not necessarily to be the pravega container, when customer inject sidecars.

### Purpose of the change

Fixes #527
Loop the containers and pick the correct container to check faulty.

### What the code does

Loop the containers and pick the correct container to check faulty.

### How to verify it

deploy Pravega and inject sidecars, and helm upgrade pravega, check if pravega can be upgraded successfully.


pkg/util/pravegacluster.go
```diff
@@ -265,10 +265,16 @@ func IsPodReady(pod *corev1.Pod) bool {
...
+}
+
 func IsPodFaulty(pod *corev1.Pod) (bool, error) {
-	if pod.Status.ContainerStatuses[0].State.Waiting != nil && (pod.Status.ContainerStatuses[0].State.Waiting.Reason == "ImagePullBackOff" ||
-		pod.Status.ContainerStatuses[0].State.Waiting.Reason == "CrashLoopBackOff") {
-		return true, fmt.Errorf("pod %s update failed because of %s", pod.Name, pod.Status.ContainerStatuses[0].State.Waiting.Reason)
+	for _, container := range pod.Status.ContainerStatuses {
+		if IsPravegaContainer(container) && container.State.Waiting != nil && (container.State.Waiting.Reason == "ImagePullBackOff" ||
+			container.State.Waiting.Reason == "CrashLoopBackOff") {
+			return true, fmt.Errorf("pod %s update failed because of %s", pod.Name, pod.Status.ContainerStatuses[0].State.Waiting.Reason)
```
anishakj, line 276: Do we need to change `pod.Status.ContainerStatuses[0]` to  `container` ?

anishakj (changes requested):

pkg/util/pravegacluster.go
realAaronWu: The first contianer may not be the pravega container so we need to loop all containers and find the pravega container

anishakj: correct, so error message has to be updated for that container

anishakj (changes requested):
Unit tests are failing, Also the last commit is not signed off. Could you please fix those

anishakj (approved):
LGTM

PR #530, opened 2021-04-27
Issue 529: Add support  to configure the service and sts suffix generated by Pravega Operator

### LINKED ISSUE #529: Shorten the service names generated by Pravega Operator
### Description

According to RFC-[1123 ](https://tools.ietf.org/html/rfc1123)and k8s [wiki](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), DNS label name should be no more than 63 characters. Currently Segment Store service name suffix already has 30 characters, which only left too little room for a customized name for Pravega apps. We need a way to shorten the suffix in the DNS labels to make rooms for customized app name.

### Importance

must-have

### Location

pravega operator and all services created by operator

shorten the service suffix and support upgrade.

anishakj:
### Change log description

Currently, in operator controller service name is hardcoded as `pravega cluster name`  followed by string `-pravega-controller`. Similarly, headless service name is  `pravega cluster name`  followed by string `-pravega-segment-store-headless` . So in case of providing  cluster names with more length installation is failing with error `service name should not be more than 63 characters`

### Purpose of the change

Fixes #529
### What the code does
Provide an option to configure service name suffix and sts name suffix.

### How to verify it

Verified the following scenarios
1. Installed Pravegacluster by configuring different suffix names for controller and segment store headless service and sts name. Service names are shown as clustername followed by configured suffix
2.  Installed pravegacluster without configuring suffix for service name and sts. Cluster is coming up with default suffix for service name and sts name
3. Verified upgrading the operator, service name and sts name is not changing and not creating new service or sts.

### DIFF
--- a/charts/pravega/templates/pravega.yaml
+++ b/charts/pravega/templates/pravega.yaml
@@ -91,6 +91,15 @@ spec:
       repository: {{ .Values.image.repository }}
       pullPolicy: {{ .Values.image.pullPolicy }}
     controllerReplicas: {{ .Values.controller.replicas }}
+    {{- if .Values.controller.svcNameSuffix }}
+    controllerSvcNameSuffix: {{ .Values.controller.svcNameSuffix }}
+    {{- end }}
+    {{- if .Values.segmentStore.stsNameSuffix }}
+    segmentStoreStsNameSuffix: {{ .Values.segmentStore.stsNameSuffix }}
+    {{- end }}
+    {{- if .Values.segmentStore.headlessSvcNameSuffix }}
+    segmentStoreHeadlessSvcNameSuffix: {{ .Values.segmentStore.headlessSvcNameSuffix }}
+    {{- end }}
     {{- if .Values.controller.maxUnavailableReplicas }}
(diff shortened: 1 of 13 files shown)


charts/pravega/templates/pravega.yaml
```diff
@@ -91,6 +91,15 @@ spec:
       repository: {{ .Values.image.repository }}
       pullPolicy: {{ .Values.image.pullPolicy }}
     controllerReplicas: {{ .Values.controller.replicas }}
+    {{- if .Values.controller.svcNameSuffix }}
```
medvedevigorek, line 94: Is it possible to set default value if no value is provided in values file?

anishakj: I can do that, but if someone tries to install pravega using this charts and old pravega operator openAPI schema will throw error as this field is not present in CRD of previous operator.

medvedevigorek: okay, let's keep it as is for now then. Thanks

medvedevigorek (approved):
Looks good to me. Thanks

SrishT (approved):
LGTM

PR #535, opened 2021-05-03
Issue 534: Provide support to configure segment store port

### LINKED ISSUE #534: Provide support to configure segment store port
### Description

Currently, segment store port is hardcoded to `12345` from operator. 

### Importance

should-have

### Location

Make the segmentstore port configurable

anishakj:

### Change log description

Provided support to configure segment store ports. If the port is not mentioned in pravega options,default port will be set as `12345`
### Purpose of the change

Fixes #534

### What the code does

Added support to configure segment store port.
### How to verify it

Verified that segment store pods are coming up with different port number, if it is configured.
If port number is not specified, segment store pod is coming up with default port number
Verified changing segment store  port at run time after Pravega is installed. 
Verified changing segment store  port at run time after Pravega is installed with external access.

### DIFF
--- a/pkg/test/e2e/e2eutil/pravegacluster_util.go
+++ b/pkg/test/e2e/e2eutil/pravegacluster_util.go
@@ -890,16 +890,18 @@ func CheckStsExists(t *testing.T, f *framework.Framework, ctx *framework.TestCtx
 	return nil
 }
 
-func CheckConfigMapUpdated(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, pravega *api.PravegaCluster, cmName string, field string) error {
+func CheckConfigMapUpdated(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, pravega *api.PravegaCluster, cmName string, key string, value string) error {
 	cm := &corev1.ConfigMap{}
 	err := f.Client.Get(goctx.TODO(), types.NamespacedName{Namespace: pravega.Namespace, Name: cmName}, cm)
 	if err != nil {
 		return fmt.Errorf("failed to obtain configmap: %v", err)
 	}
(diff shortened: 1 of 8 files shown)


gaddamas (commented):
Overall this looks good, however when customizing port in single IP environment (multiple segmentstores exposed on single IP with different ports) customer should be aware they still need to open-up ports on firewall, since we take configured port and increment for each additional ss replica.

pkg/test/e2e/e2eutil/pravegacluster_util.go
```diff
@@ -885,9 +886,25 @@ func CheckStsExists(t *testing.T, f *framework.Framework, ctx *framework.TestCtx
...
 	return nil
 }
 
+func CheckConfigMapUpdated(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, pravega *api.PravegaCluster, cmName string, field string) error {
+	cm := &corev1.ConfigMap{}
+	err := f.Client.Get(goctx.TODO(), types.NamespacedName{Namespace: pravega.Namespace, Name: cmName}, cm)
+	if err != nil {
+		return fmt.Errorf("failed to obtain configmap: %v", err)
+	}
+	javaopts := cm.Data["JAVA_OPTS"]
```
SrishT, line 899: there can also be a check to see if the configmap is `nil` and to check whether it contains the provided key or not. Also the key being passed can be another argument passed to the method to make it more flexible.

SrishT (changes requested):

pkg/test/e2e/e2eutil/pravegacluster_util.go
anishakj: Done

SrishT (approved):
LGTM

Issue #460, opened 2020-10-06
Include PrintExtendedThreadInfo to the list of default JVM options 


medvedevigorek:
### Description

Since JRE11 additional fields are printed for each in a thread dump. These fields are `cpu`, `elapsedTime`, `allocated` and `defined_classes`.

Only first two are printed by default, the other fields must be explicitly enabled by a JVM option 

More details can be found here -> https://bugs.openjdk.java.net/browse/JDK-8200720 
### Importance

This additional information is useful during the service actions when a thread dump is generated 

### Location

Charts/values files 

### Suggestions for an improvement

Include  `-XX:+PrintExtendedThreadInfo` to the list of default JVM options for both Controller and SegmentStore

anishakj:
I think we have to add this option in charts while we publish pravega  `0.9.0` . This JVM option should not be specific to operator version.

medvedevigorek:
Agree.  Ideally we shouldn't have many JVM options hard coded in the operator code and have them provided via the charts as it makes the whole solution more flexible. At the same time we need to set the default flags in charts

Ranganaths8:
@medvedevigorek JVM options are customizable on operator side.. We do set default values for this parameters.

anishakj:
Fixed by PR https://github.com/pravega/pravega-operator/pull/546

PR #546, opened 2021-05-16
Issue 537: Making JVM Options configurable

### LINKED ISSUE #537: JVM options should not be hardcoded within the operator
### Description
Currently many JVM options are all hardcoded within the operator, which does not allow users to remove or modify the options added by the operator by default, even if they wish to do so.

### Importance
must have

### Location
pkg/controller/pravega/pravega_controller.go

The recommended JVM options along with their default values should be specified within the values.yaml file instead, so that the user can easily modify them if required.

SrishT:

### Change log description
Removes the hardcoded JVM options from the operator code, so that users can easily set their desired configuration through the charts, and also modify them if required.

### Purpose of the change
Fixes #537, #460 

### What the code does
Removes the hardcoded JVM options from the operator code.

### How to verify it
The segmentstore and controller should be configured with the JVM options that are provided within the values.yaml by the user. Also these values can be easily modified even after the pravega cluster has been deployed by editing the pravega cluster spec.

### DIFF
--- a/pkg/test/e2e/e2eutil/pravegacluster_util.go
+++ b/pkg/test/e2e/e2eutil/pravegacluster_util.go
@@ -890,25 +890,21 @@ func CheckStsExists(t *testing.T, f *framework.Framework, ctx *framework.TestCtx
 	return nil
 }
 
-func CheckConfigMapUpdated(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, pravega *api.PravegaCluster, cmName string, key string, value string) error {
+func CheckConfigMapUpdated(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, pravega *api.PravegaCluster, cmName string, key string, values []string) error {
 	cm := &corev1.ConfigMap{}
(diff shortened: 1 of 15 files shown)


pkg/test/e2e/e2eutil/pravegacluster_util.go
```diff
@@ -890,21 +890,21 @@ func CheckStsExists(t *testing.T, f *framework.Framework, ctx *framework.TestCtx
...
 		optvalue := cm.Data[key]
-		if strings.Contains(optvalue, value) {
-			t.Logf("configmap is updated %s", cm.Name)
-			return nil
+		for _, value := range values {
+			if !strings.Contains(optvalue, value) {
+				return fmt.Errorf("config map is not updated")
+			}
 		}
 	}
```
anishakj, line 906: can we return error in case of config map is not found

anishakj (changes requested):

pkg/test/e2e/e2eutil/pravegacluster_util.go
SrishT: done

anishakj (approved):
LGTM

PR #556, opened 2021-06-08
Issue 483: Auth handler implementation using init containers

### LINKED ISSUE #484: Enhancement request to support init containers for Pravega
### Description

The Pravega controller supports a pluggable interface `AuthHandler` which allows implementers to provide their own authentication and authorization implementation.

To enable a Pravega controller process to load a custom implementation of the interface, one simply has to drop in the corresponding jar in `/opt/..../plugins`.

However that simple task tends to be somewhat unnecessarily complicated in the Kubernetes/docker world, since that means the implementer has to build a new docker image, based on the Pravega stock image, to include the new library.   

### Importance

The above complicates iterative development of the plugin implementation and forces non stock Pravega images to exist.

### Location

N/A

A more efficient way could be the following:   provide a way to supply an init-container image to the `PravegaCluster` resource.   The operator would pick up on that and inject the init container image into the pravega controller Deployment, for the purpose of transferring jars (a common pattern with init containers in K8).

The operator would then insure to copy the jar librarie(s) from the initcontainer to an `emptyDir` volume type, then from the `emptyDir` to the plugin directory, as part of the normal lifecycle of the Pravega controller pod.

Something very similar was done for the Schema Registry project chart, which also uses the `AuthHandler` interface.

For Pravega proper, it would have to be allowed in the `PravegaCluster` spec and handled by the Pravega operator.

To ease the injection of the plugin implementation even further, we could make the integration of the init container even more specific and guided than what the Schema registry did.  In other words, The proposal would be for the `PravegaCluster` spec to have a new field along the lines of:

```
authImplementations:
-image1:1.0
-image2:5.3
...
```

The responsibility of the implementer of the supplied image(s) would be to provide a docker image which contains in a known location (say /plugins/implementation/) a valid jar with the implementation of the `AuthHandler` interface.

At deployment time, the Pravega operator would then:
1.  cycle through the `authImplementations` list, and insert as many init containers as there are images, in the controller Deployment.  For each initcontainer of the controller pod, it would provide the command to execute, which would be `cp /plugins/implementation/*.jar /some-mapped-volume-based-on-emptydir-type`).

2. Next, the operator would map that same `emptyDir` volume into the controller pod, under some location, say /some-well-known-source-dir-for-plugins.

3. Next, the Pravega image itself, would need, in its startup script, a pre command to copy the contents of /some-well-known-source-dir-for-plugins into `/opt/.../plugins`.

Alternatively (to 3 above), it could potentially also map the `emptyDir` volume, (now loaded with the jar files) directly under /opt/..../plugins if that is possible.

**What this achieves:**

As an implementer of the `AuthHandler` interface, I can simply keep patching the `authImplementations` field of the `PravegaCluster` resource and have the operator automatically restart my controller with the new libraries in place.

That would greatly decouple Pravega proper from its plugin implementations at runtime.

anishakj:

### Change log description

If Authhandler is speecified, init container is created and it will copy the binary and mount to an emptydir

### Purpose of the change

 Fixes #484

### What the code does

_(Detailed description of the code changes)_

### How to verify it

Verified that able to copy jar files to the controller pod


maddisondavid:
I'm wondering if this could be achieved with just the standard generic init-container support and some documentation.  The problem is that there could be other artifacts required such as a Bookkeeper Rack Aware placement script which would also need to be copied.

My suggestion would be that this use case is handled by the user with an init container image.  The Operator would mount an `EmptyDir` at a well known documented path, i.e. `/mnt/pravega` into which the init container could copy any required artifacts.  Documentation could provide a sample `Dockerfile` and bash script.

sarlaccpit:
@maddisondavid , @anishakj 

Ok this is a very encouraging PR.   Let me pipe in a couple details; some of which were discussed offline, so it'd be best to share here.

Generic init containers (for controller, segment store pods) that literally are straight up init containers out of the K8 spec.  VS. First class specialized pieces of the PravegaCluster spec that are known to the Operator it knows exactly what to do with them(maybe by doing init container stuff underneath, maybe not) 

In the first case, the operator knows nothing at all and just appends the initcontainer to the deployment/statefulsets.   In the second, the operator knows exactly what it's dealing with, it knows which pod it actually is relevant to(ctrl vs ss) and where to put what where.   

Both have value and honestly my opinion is why not have both?  Known use cases that Pravega supports in terms of customization, why not first class them?   AND also have generic init containers for whatever we haven't thought about yet.

Why have specialized support?   Because, look at these 2 use cases:
- **auth plugin implementation**: do we want the user to have to write his/her own init container spec to copy the file to a emptyDir volume, and know where to mount it in the final target location?  And the user has to know that in this case, it concerns the controller, not the SS.
- **bk placement awareness script**:   same requirement. Except this time, the user has to know it concerns the SS.

That sounds like a a bit too much to ask the user to know/learn, even with documentation.  (maybe).

By first classing the two above, you ask the user for:
- auth plugin: an image that has a jar in it in some defaulted location.  Done.  Operator does the rest and instruments the controller pods automatically.
- bk placement script: an image that has an executable script in it in some location.   Done. Operator does the rest and instruments the SS pods automatically.

In this PR, @anishakj is going for some generic initcontainer support and a first class support for auth plugins.

charts/pravega/templates/pravega.yaml
```diff
@@ -111,6 +111,10 @@ spec:
       repository: {{ .Values.image.repository }}
       pullPolicy: {{ .Values.image.pullPolicy }}
     controllerReplicas: {{ .Values.controller.replicas }}
+    {{- if .Values.segmentStore.initContainers }}
+    authImplementations:
+{{ toYaml .Values.authImplementations | indent 6 }}
```
sarlaccpit, line 116: I'm not following the logic why the authImplementations is only added in if segmentStore.initContainers is not empty.

charts/pravega/values.yaml
```diff
@@ -97,7 +97,10 @@ segmentStore:
   stsNameSuffix:
   headlessSvcNameSuffix:
   initContainers: []
-
+authImplementations: []
+# - image:
+#   pluginLocation: "/sdp/data/*"
```
sarlaccpit, line 102:
```suggestion
#   pluginLocation: "/some_vendor/data/*"
```

pkg/controller/pravega/pravega_controller.go
```diff
@@ -206,8 +208,53 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
...
+				Name:    "authplugin" + strconv.Itoa(i),
+				Image:   plugin.Image,
+				Command: []string{"/bin/sh"},
+				Args: []string{
+					"-cx",
+					arg,
+				},
+				VolumeMounts: []corev1.VolumeMount{
+					{
+						Name:      "authplugin" + strconv.Itoa(i),
```
sarlaccpit, line 242: I believe the mountPath is static (and well known to the pravega controller), and there are, since we're in a loop, X volumeMounts getting mounted to the same mount path.    

Is that a problem (is one masking the other) or are they merged somehow (with /opt/pravega/pluginlib/  showing an aggregation of all the mounted "files").

pkg/controller/pravega/pravega_controller.go
```diff
@@ -206,8 +208,53 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
 	if p.Spec.Pravega.ControllerSecurityContext != nil {
 		podSpec.SecurityContext = p.Spec.Pravega.ControllerSecurityContext
 	}
+	podSpec.InitContainers = []corev1.Container{}
 	if p.Spec.Pravega.ControllerInitContainers != nil {
-		podSpec.InitContainers = p.Spec.Pravega.ControllerInitContainers
+		podSpec.InitContainers = append(podSpec.InitContainers, p.Spec.Pravega.ControllerInitContainers...)
```
sarlaccpit, line 213: Say I am writing my own initcontainer and I add in there a command to copy some things in a volumeMount path.   But I'd have to refer to a volume.  That volume is at the pod level.   That volume is only created in the case of the AuthImplementations, atm.

Does that mean the generic initcontainers as they are right now, would be essentially useless?  (from the pov of copying something in the main container I mean.  I suppose init containers can have other uses as well)

charts/pravega/templates/pravega.yaml
anishakj: it was copy/paste error. Fixed that

charts/pravega/values.yaml
anishakj: Changed

pkg/controller/pravega/pravega_controller.go
anishakj: yes, it is an issue. mount path should be unique.   Modified the code to have unique mount path

pkg/controller/pravega/pravega_controller.go
anishakj: Yes, in addition to init containers, we have to create empty dir volume. This can be achieved by providing `emptyDirVolumeMounts` in pravega options

charts/pravega/README.md
```diff
@@ -158,7 +158,8 @@ The following table lists the configurable parameters of the pravega chart and t
 | `segmentStore.headlessSvcNameSuffix` | suffix for segmentStore headless service name | `pravega-segmentstore-headless` |
 | `segmentStore.labels` | Labels to add to the segmentStore pods | `{}` |
 | `segmentStore.annotations` | Annotations to add to the segmentStore pods | `{}` |
-| `segmentStore.initContainers` | Init Conatiners to add to the segmentStore pods | `[]` |
+| `segmentStore.initContainers` | Init Containers to add to the segmentStore pods | `[]` |
+| `authImplementations` | Plugin configuration to be used | `[]` |
```
sarlaccpit, line 162: I think the README could use a tiny bit more guidance for:
- the user interested in the the authImplementation (the sample values.yaml is very helpful, but mentioning here that the user needs to provide an image with that implementing the AuthHandler (https://github.com/pravega/pravega/blob/master/documentation/src/docs/auth/auth-plugin.md); and the image should have /bin/sh available (sounds obvious, but it's not guaranteed, depending how the user crafts his/her Dockerfile.)
- for the generic controller segment store initcontainers, mention that it can be used in conjunction of the emptyDirVolume option you were mentioning elsewhere.

Calling all this out up front in the README will be very helpful.

anishakj: Updated the documentation

doc/auth-handlers.md
```diff
@@ -0,0 +1,34 @@
+## AuthImplementations
+
+If we need to enable auth handlers in Pravega controller, modify the pravega manifest and ensure that following fields are present in the yaml file.
+
+```
+authImplementations:
+  mountPath: "/opt/pravega/pluginlib"
+  authHandlers:
+ - image: authHandlerImage
```
sarlaccpit, line 9: slight misaligned indent.

doc/auth-handlers.md
```diff
@@ -0,0 +1,34 @@
+## AuthImplementations
+
+If we need to enable auth handlers in Pravega controller, modify the pravega manifest and ensure that following fields are present in the yaml file.
```
sarlaccpit, line 3: second pravega is not capitalized.

sarlaccpit (commented):
Minor nits on the read me. Thanks for adding that.

Unrelated:  just curious how this plays out with upgrade?   I don't think it should be a problem but I was just wondering if you considered/tested it.

For example, a PravegaCluster resource that does not have the initContainer or authImplementation enabled.     We then upgrade the pravega operator and the resource to have both these things enabled.

It should just work I believe?  Could we make sure?

doc/auth-handlers.md
anishakj: Fixed

doc/auth-handlers.md
anishakj: Fixed , Also verified the upgrade use case. It is working fine

pkg/controller/pravega/pravega_controller.go
```diff
@@ -206,8 +207,54 @@ func makeControllerPodSpec(p *api.PravegaCluster) *corev1.PodSpec {
...
+		if p.Spec.Pravega.AuthImplementations.MountPath != "" {
+			mountPath = p.Spec.Pravega.AuthImplementations.MountPath
+		} else {
+			mountPath = "/opt/pravega/pluginlib"
+		}
+		for i, plugin := range p.Spec.Pravega.AuthImplementations.AuthHandlers {
+			if plugin.Source != "" {
+				srcPath = plugin.Source
+			} else {
+				srcPath = "/plugins/implementation/*.jar"
```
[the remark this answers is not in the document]
anishakj: @sarlaccpit should i make it mandatoy?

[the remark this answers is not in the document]
sarlaccpit: @anishakj ; it's a good question.  I don't feel very strongly about it along as the documentation on the behavior is very crisp on what is expected.

Having a default location means it's one less thing the user needs to provide, HOWEVER, it has to be very clear to that user that /plugins/implementations/ is where he/she's expected to put the jar(s).      It's true that if you source the srcPath to be required, you are putting that explicitly in front of the user...

Honestly, I don't have a preference.  :)  You pick according to what you feel is best; and please document accordingly.

doc/auth-handlers.md
```diff
@@ -0,0 +1,34 @@
+## AuthImplementations
+
+If we need to enable auth handlers in Pravega controller, modify the Pravega manifest and ensure that following fields are present in the yaml file.
+
+```
+authImplementations:
+  mountPath: "/opt/pravega/pluginlib"
```
sarlaccpit, line 7: One thing I was wondering about mountPath.   SHOULD it be something the user provides?   What I mean is, if they change it, things will actually not work (because I think the controller looks specifically in that location).    If that's true, then it shouldn't be configurable at all I feel.

anishakj: I felt that making it  configurable is better. If the path in controller changes later, from operator side no changes will be required

sarlaccpit: got it.

doc/auth-handlers.md
```diff
@@ -0,0 +1,34 @@
...
+  - image: authHandlerImage
+    source: "/some_vendor/data/*"
+```  
+
+User has to provide an image that contains authhandler implementation and the image should have `/bin/sh` enabled.
+
+Pravega performs dynamic implementations of the Authorization/Authentication [API](https://github.com/pravega/pravega/blob/master/documentation/src/docs/auth/auth-plugin.md). Multiple implementations can co-exist and different plugins can be used by different users.
+
+
+Alternatively, user can mention the details in init Containers as well to copy the files to controller or segmentstore pods. For using init containers modify the  Pravega manifest and ensure that following fields are present.
```
sarlaccpit, line 18: I think the section about the generic init containers should come outside of the auth-handlers.md because it's not specific to auth.

In fact, one of the very real use cases for the generic init containers it the copying of the bk placement script into the segment store container.    So it should probably be placed in its own .md file.

sarlaccpit (approved):
Thank you for taking the time to provide the extra documentation.

SrishT (approved):
LGTM

PR #561, opened 2021-07-01
Issue 557: Many Pravega Spec parameters are still not updatable at runtime

### LINKED ISSUE #557: Many Pravega Spec parameters are still not updatable at runtime
### Description
There are many Pravega Spec parameters still remaining which cannot be updated once the pravega cluster has been created. All the spec fields other than the the version, replica count, DNS name, JVM options as well as the pravega options still come under this category. While not allowing updates on some parameters might make sense, it will be necessary to allow updates to a few others including controllerResources, segmentStoreResources, segmentStoreEnvVars, segmentStoreSecret etc.  

### Importance
must have

### Location
pkg/controller/pravegacluster/pravegacluster_controller.go

The operator reconcile logic would need to be modified so that updates to necessary fields are allowed during the lifetime of the cluster. 
A maintenance mode can be added to stop operator from watching all resources and allow the user to make any changes to recover the cluster.

anishakj:
### Change log description

Currently many parameters belonging to sts and deployment are not updatable at runtime. eg; Resources

### Purpose of the change

 Fixes #557

### What the code does

Made changes for updating statefulset and deployment if there are any changes to them

### How to verify it

Modified pravega cluster fields like resources and verified that sts and deployments are getting updated and pods are getting restarted.

### DIFF
--- a/pkg/controller/pravegacluster/pravegacluster_controller.go
+++ b/pkg/controller/pravegacluster/pravegacluster_controller.go
@@ -635,10 +635,7 @@ func (r *ReconcilePravegaCluster) deployController(p *pravegav1beta1.PravegaClus
 			return err
 		}
 
-		deployment.Spec.Template.Spec.Containers[0].Image = foundDeploy.Spec.Template.Spec.Containers[0].Image
-		deployment.Spec.Template.Annotations["pravega.version"] = foundDeploy.Spec.Template.Annotations["pravega.version"]
-		//if !r.checkVersionUpgradeTriggered(p) {
-		if !reflect.DeepEqual(deployment.Spec.Template, foundDeploy.Spec.Template) {
+		if !r.checkVersionUpgradeTriggered(p) && !r.isRollbackTriggered(p) {
 			foundDeploy.Spec.Template = deployment.Spec.Template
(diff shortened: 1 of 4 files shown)


pkg/controller/pravegacluster/pravegacluster_controller.go
```diff
@@ -650,19 +668,23 @@ func (r *ReconcilePravegaCluster) deploySegmentStore(p *pravegav1beta1.PravegaCl
...
 				err = r.client.Update(context.TODO(), sts)
 				if err != nil {
 					return fmt.Errorf("failed to update stateful set: %v", err)
 				}
-				err = r.restartStsPod(p)
-				if err != nil {
-					return err
+
+				if !reflect.DeepEqual(originalsts.Spec.Template, sts.Spec.Template) {
+					err = r.restartStsPod(p)
```
SrishT, line 681: `restartStsPod` will introduce a 10 minute delay if there is an issue with any pod in coming back to ready state, could we have a workaround for that?

anishakj: Raised https://github.com/pravega/pravega-operator/issues/562 to track this

SrishT (approved):
LGTM

PR #563, opened 2021-07-14
Issue 280: Read InfluxDB credentials from a secret

### LINKED ISSUE #280: Read InfluxDB credentials from a secret
### Description

Right now the only way to pass influxdb credentials to pravega (needed to store pravega metrics) is via `pravega_options` block of the custom `pravegacluster` resource. 
As a result the above credentials are hanging around in plain text and are available via `kubectl get pravegacluster -o yaml` command.
It would be much better if pravega-operator had an option to take a name of the influxdb secret that holds the credentials (assuming that influxdb is deployed in the same namespace as pravega) and read the influxdb credentials from there.

### Importance

Having the influxdb credentials in plain text in the CR poses significant security risk.

### Location

https://github.com/pravega/pravega-operator/blob/master/pkg/apis/pravega/v1alpha1/pravega.go 

Add a new `influxdb_secret` spec to pravega spec, that would take the name of influxdb secret, read database credentials from the secret, then create 
- `controller.metrics.influxDBUserName` and `controller.metrics.influxDBPassword` options for pravega controller 
- `metrics.influxDBUserName` and `metrics.influxDBPassword` options for pravega segmentstore

anishakj:

### Change log description

Added a new field `InfluxDBSecret` in  Pravega spec, so that users can mention the secret names  rather than passing username/password as plain text.

### Purpose of the change

Fixes #280

### What the code does

It will read the secret from the user, and if it is mentioned secret is mounted in controller and segment store pods at `/etc/influxdb-secret-volume/`

### How to verify it

Verified that secret is mounted at controller and segment store pods.


pkg/controller/pravega/pravega_controller.go
```diff
@@ -276,6 +277,20 @@ func configureControllerAuthSecrets(podSpec *corev1.PodSpec, p *api.PravegaClust
 	}
 }
 
+func configureControllerInfluxDBSecrets(podSpec *corev1.PodSpec, p *api.PravegaCluster) {
+	if p.Spec.Pravega.InfluxDBSecret != "" {
+		addSecretVolumeWithMount(podSpec, p, influxDBSecretVolumeName, p.Spec.Pravega.InfluxDBSecret,
```
SrishT, line 282: could  we consider accepting the mount path as a property from the user instead of hardcoding its value in a constant?

SrishT (changes requested):

SrishT (approved):
LGTM

PR #565, opened 2021-07-14
Issue 564: Removing supported versions configmap

### LINKED ISSUE #564: Supported versions configmap should be removed
### Description
The operator currently validates the version of the pravega image before installation, and also checks the version before allowing upgrade or rollback, using the validating webhook. While we still want to only allow rollback to the last stable version before the upgrade failed, we want to remove the version validation which is currently done before any installation or upgrade, and hence maintaining the supported versions configmap should no longer be necessary.

### Importance
nice to have

### Location
charts/pravega-operator/templates/version_map.yaml
deploy/version_map.yaml

SrishT:

### Change log description
Removes the supported versions configmap from the pravege operator.

### Purpose of the change
Fixes #564 

### What the code does
Removes the version validation from the validating webhook at the time of installation or upgrades of the pravega cluster. However the webhook will continue to check the format of the image being provided and will not allow any downgrade, unless while trying to rollback the cluster in case of a failed upgrade.

### How to verify it
The pravega operator should allow installation of any pravega version given in the correct format (i.e. *MAJOR*.*MINOR*.*PATCH*-*XXXX*), and should also allow upgrades to any other pravega version. Downgrades are however not allowed.

The following scenarios have been tested
- Upgrading the pravega operator from version 0.5.4 (which contains the version map) to the latest one (not containing the version map)
- Tested installation and upgrade to any pravega version is allowed
- Tested that the pravega operator does not allow using images which do not follow semantic versioning, and also does not allow the user from downgrading the pravega version

### DIFF
--- a/pkg/apis/pravega/v1beta1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1beta1/pravegacluster_types.go
@@ -972,7 +972,7 @@ func (p *PravegaCluster) ValidatePravegaVersion() error {
 	if match, _ := util.CompareVersions(normRequestVersion, normFoundVersion, "<"); match {
 		return fmt.Errorf("downgrading the cluster from version %s to %s is not supported", p.Status.CurrentVersion, requestVersion)
 	}
-	log.Printf("validateBookkeeperVersion:: normFoundVersion %s", normFoundVersion)
+	log.Printf("ValidatePravegaVersion:: normFoundVersion %s", normFoundVersion)
 
 	log.Print("ValidatePravegaVersion:: No error found...returning...")
 	return nil
--- a/pkg/apis/pravega/v1beta1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1beta1/pravegacluster_types.go
@@ -11,7 +11,6 @@
 package v1beta1
 
 import (
-	"bufio"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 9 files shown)


pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1009,16 +969,11 @@ func (p *PravegaCluster) ValidatePravegaVersion(filename string) error {
...
-		// It should never happen
-		return fmt.Errorf("failed to find current cluster version in the supported versions")
-	}
-	upgradeList := strings.Split(upgradeString, ",")
-	if !util.ContainsVersion(upgradeList, normRequestVersion) {
-		return fmt.Errorf("unsupported upgrade from version %s to %s", p.Status.CurrentVersion, requestVersion)
+	if match, _ := util.CompareVersions(normRequestVersion, normFoundVersion, "<"); match {
+		return fmt.Errorf("downgrading the cluster from version %s to %s is not supported", p.Status.CurrentVersion, requestVersion)
 	}
+	log.Printf("validateBookkeeperVersion:: normFoundVersion %s", normFoundVersion)
```
anishakj, line 975: Change to PravegaVersion

anishakj (changes requested):
Remove all the references of version map from documentation.

pkg/apis/pravega/v1beta1/pravegacluster_types.go
SrishT: done

anishakj (approved):
LGTM

PR #586, opened 2021-10-21
Issue 495: Validate Pravega manifest settings before deployment

### LINKED ISSUE #495: Validate Pravega manifest settings before deployment.
### Description

https://github.com/pravega/pravega-operator/blob/master/doc/upgrade-cluster.md#upgrade-to-pravega-07-or-above indicates that Segment store settings following rules like : 

POD_MEM_LIMIT = JVM Heap + Direct Memory
Direct Memory = pravegaservice.cache.size.max + 1GB/2GB (other uses) ; (Direct memory is less than cache.size.max)

This verification can be done before deployment by the operator instead of manually relying on the user to set it right...

With incorrect values, the Segment store faces errors of the pattern.
```
 00:39:19,662 10936089 [core-23] ERROR i.p.s.server.logs.DurableLog - DurableLog[40] Recovery FAILED.
io.pravega.segmentstore.storage.cache.CacheFullException: DirectMemoryCache full: Stored = 11788056931, Used = 11811160064, Reserved = 23068672, Allocated = 11811160064, Max = 11811160064
```
### Importance

should-have

### Location

Pre deployment validation..

Ensure Pre-condition checks like the one mentioned for SegStore are performed before deployment.

nishant-yt:
Signed-off-by: Nishant Gupta <Nishant_Gupta3@dell.com>

### Change log description

While deploying Pravega , Segment Store settings should adhere to the following rules:
POD_MEM_LIMIT > JVM Heap (`-Xmx`) + JVM Direct Memory (`-XX:MaxDirectMemorySize`)
JVM Direct memory > Segment Store read cache size (`pravegaservice.cache.size.max`)

Here we are validating the above scenarios before deployment by the operator happens instead of manually relying on the user to set them right.
### Purpose of the change

Fixes #495

### What the code does

The code makes sure the following checks are performed by the operator before deploying Pravega:
POD_MEM_LIMIT > JVM Heap + JVM Direct Memory
JVM Direct memory > Segment Store read cache size

### How to verify it

Try setting different values for` .spec.pravega.segmentStoreResources.requests.memory` , `-Xmx` , `-XX:MaxDirectMemorySize` and `pravegaservice.cache.size.max` and the operator should return error in case the above mentioned rules are not met.


pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,66 @@ func (p *PravegaCluster) validateConfigMap() error {
 	return nil
 }
 
+func (p *PravegaCluster) ValidateSegmentStore() error {
+	totalMemoryLimitsQuantity := p.Spec.Pravega.SegmentStoreResources.Limits[corev1.ResourceMemory]
+	totalMemoryRequestsQuantity := p.Spec.Pravega.SegmentStoreResources.Requests[corev1.ResourceMemory]
```
anishakj, line 1211: How are you verifying the scenario, limit is set but requests are not set for resources

pkg/test/e2e/e2eutil/spec_util.go
```diff
@@ -33,14 +33,43 @@ func NewDefaultCluster(namespace string) *api.PravegaCluster {
...
+				SegmentStoreResources: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("1000m"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("2000m"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+					},
+				},
```
jkhalack, line 49: Where do these values come from?

jkhalack: Setting all the defaults here without verifying user supplied values is extremely dangerous.

If the user supplies memory limit less than `4Gi` (say, `2Gi`) without memory request,
they will end with a `4Gi` memory request (default) and `2Gi` memory limit (user-supplied).

What we rather need to do while setting the defaults is:
1. See if the user has set a memory request
2. If the request is not set, then set it to memory limit (the admission webhook would be verifying that the memory limit is set by the user).

pkg/apis/pravega/v1beta1/pravegacluster_types.go
jkhalack: Setting the request to the limit value should be done in the defaults section, if I am not mistaken.

@nishant-yt Did you verify that `totalMemoryRequestsQuantity.Value()` below produces a zero when the memory request is NOT set by the user?

nishant-yt: Added the code in defaults section to make sure requests is equal to limits in case requests is not set by the user. Also,  I have verified `totalMemoryRequestsQuantity.Value()` does produces `0` value if memory requests is not set by the user

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,66 @@ func (p *PravegaCluster) validateConfigMap() error {
 	return nil
 }
 
+func (p *PravegaCluster) ValidateSegmentStore() error {
```
RaulGracia, line 1209: Please, add comments in this method explaining what it does and why. Also, the name of the method should be more concrete, like `ValidateSegmentStoreMemorySettings`, as `ValidateSegmentStore` is not really descriptive about what it is doing.

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,66 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+	if maxDirectMemoryString == "" {
+		return fmt.Errorf("Missing required value for Segment Store JVM option -XX:MaxDirectMemorySize")
+	}
+	maxDirectMemoryQuantity := resource.MustParse(maxDirectMemoryString)
+
+	xmx := xmxQuantity.Value()
+	maxDirectMemorySize := maxDirectMemoryQuantity.Value()
+	cacheSize := cacheSizeQuantity.Value()
+
+	if totalMemoryLimits <= maxDirectMemorySize+xmx {
```
RaulGracia, line 1254: Nit: please, leave a space after and before the `+` symbol (it is hard to read otherwise).

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,66 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+		return fmt.Errorf("MaxDirectMemorySize(%v B) along with JVM Xmx value(%v B) is greater than or equal to the total available memory(%v B)!", maxDirectMemorySize, xmx, totalMemoryLimits)
+	}
+
+	if maxDirectMemorySize <= xmx {
```
RaulGracia, line 1258: Please, remove this part of the validation (also from the PR description), as this may be completely legitimate. You may want to configure a JVM of 4GB with a cache of 3GB (and therefore setting `maxDirectMemorySize=4GB`). And there is nothing strictly wrong with that, is just a user choice. We have to validate to prevent what we know for sure is wrong, but not more than that, as we also want to give some degree of flexibility to users.

RaulGracia (changes requested):

pkg/apis/pravega/v1beta1/pravegacluster_types.go
nishant-yt: According to this [document](https://github.com/pravega/pravega/blob/master/documentation/src/docs/admin-guide/segmentstore-memory.md), we should generally provision the Segment Store with more direct memory than heap memory. So, shouldn't this validation be there. I agree with your example and can remove the `equality` from the check so as to provide flexibility to the user.

anishakj:
@nishant-yt Could you please add end to end tests to cover the webhook validations?

pkg/test/e2e/e2eutil/spec_util.go
nishant-yt: This default values are only for e2e tests and the checks are already in place for all the scenarios

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,71 @@ func (p *PravegaCluster) validateConfigMap() error {
...
 }
 
+// ValidateSegmentStoreMemorySettings checks whether the user has passed the required values for segment store memory settings.
+// The required values includes: SegmentStoreResources.limits.memory, pravegaservice.cache.size.max , -XX:MaxDirectMemorySize and -Xmx.
+// Once the required values are set, the method also checks whether the following conditions are met:
+// Total Memory > JVM Heap (-Xmx) + JVM Direct Memory (-XX:MaxDirectMemorySize)
+// JVM Direct memory > Segment Store read cache size (pravegaservice.cache.size.max).
+func (p *PravegaCluster) ValidateSegmentStoreMemorySettings() error {
+	if p.Spec.Pravega.SegmentStoreResources == nil {
+		return fmt.Errorf("Missing required value for field spec.pravega.segmentStoreResources.limits.memory")
```
jkhalack, line 1216: @RaulGracia Should we allow deployments without setting the resources?

pkg/apis/pravega/v1beta1/pravega.go
```diff
@@ -323,6 +323,21 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
...
+
+	if s.SegmentStoreResources.Limits[v1.ResourceCPU] == (resource.Quantity{}) {
+		changed = true
+		s.SegmentStoreResources.Limits[v1.ResourceCPU] = resource.MustParse(DefaultSegmentStoreLimitCPU)
+	}
+
+	if s.SegmentStoreResources.Requests[v1.ResourceCPU] == (resource.Quantity{}) {
+		changed = true
+		s.SegmentStoreResources.Requests[v1.ResourceCPU] = resource.MustParse(DefaultSegmentStoreRequestCPU)
+	}
```
jkhalack, line 339: I do not think this is a good idea.
If the user sets the CPU request and is not aware of the default limit, we can create a situation where the limit is below the request.

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,71 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+	}
+
+	totalMemoryLimitsQuantity := p.Spec.Pravega.SegmentStoreResources.Limits[corev1.ResourceMemory]
```
jkhalack, line 1219: We also need a check that `p.Spec.Pravega.SegmentStoreResources.Limits != nil` and that `p.Spec.Pravega.SegmentStoreResources.Requests != nil` before we go ahead and check for memory settings (the motivation being that the user might set requests without limits or limits without requests).

jkhalack (changes requested):

pkg/apis/pravega/v1beta1/pravegacluster_types.go
jkhalack: Following the offline discussion, we do need a check for `p.Spec.Pravega.SegmentStoreResources` not to be `nil`.
If we allow the `nil` object here, skip the below validation and let the defaults be assigned later by [`withDefaults()` function](https://github.com/pravega/pravega-operator/blob/b5c2920839fb0b976ba49b1f8d6b495adc373ef9/pkg/apis/pravega/v1beta1/pravega.go#L312), we cannot perform a check for user-supplied `pravegaservice.cache.size.max` value or JVM memory settings. Therefore it would make sense to make `p.Spec.Pravega.SegmentStoreResources` a required block.

pkg/apis/pravega/v1beta1/pravega.go
RaulGracia: We should always ask for the memory `limit`, while the `request` may be optional. In case of setting only `limit`, `request` can be set to `limit`, as this seems to be the behavior of Kubernetes: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

jkhalack: @RaulGracia What about CPU?

pkg/apis/pravega/v1beta1/pravegacluster_types.go
RaulGracia: I agree that `limits` should not be `null`. We could leave `requests` nullable, and if so, set `requests=limits` as it seems to be the underlying behavior in Kubernetes itself.

pkg/apis/pravega/v1beta1/pravega.go
RaulGracia: While I think that CPU is less of a problem, it may be good also bound the CPU resources similarly to what is being done here for memory. In this case, we should i) enforce that at least `limit` on CPU is set (to prevent a busy or malfunctioning pod starving all other pods in the node), ii) if `requests` is not set, default it to `limit`, and iii) if `requests` is set, check that it is <= `limit`.

pkg/apis/pravega/v1beta1/pravegacluster_types.go
jkhalack: There still needs to be a check for a `nil` `Requests` object before we try to assign `Requests[corev1.ResourceMemory]` to anything (like we currently do it on the next line)

pkg/apis/pravega/v1beta1/pravega.go
nishant-yt: @jkhalack actually I will be modifying the `withDefaults()` method as shown below  so as to have the same check what we have for memory in addition to the checks what @RaulGracia mentioned  :
```
if s.SegmentStoreResources.Requests[v1.ResourceCPU] == (resource.Quantity{}) {
		changed = true
		s.SegmentStoreResources.Requests[v1.ResourceCPU] = s.SegmentStoreResources.Limits[v1.ResourceCPU]
	}
```

pkg/apis/pravega/v1beta1/pravegacluster_types.go
nishant-yt: As discussed , I will be adding a `nil` check for both `Requests` and `Limits` Objects.

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1194,6 +1206,93 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+		return fmt.Errorf("spec.pravega.segmentStoreResources cannot be empty")
+	}
+
+	if p.Spec.Pravega.SegmentStoreResources.Limits == nil {
+		return fmt.Errorf("spec.pravega.segmentStoreResources.limits cannot be empty")
+	}
+
+	if p.Spec.Pravega.SegmentStoreResources.Requests == nil {
+		return fmt.Errorf("spec.pravega.segmentStoreResources.requests cannot be empty")
+	}
```
jkhalack, line 1225: The PR changes in `pravega.go` do allow for the requests not being set.
The better way to deal with the situation is to set `totalMemoryRequests` to zero if `segmentStoreResources.requests` is not set (otherwise proceed normally).

jkhalack (approved):
LGTM

pkg/apis/pravega/v1beta1/pravega.go
```diff
@@ -323,6 +323,24 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
 		}
 	}
 
+	if s.SegmentStoreResources.Requests == nil {
+		changed = true
+		s.SegmentStoreResources.Requests = map[corev1.ResourceName]resource.Quantity{
```
RaulGracia, line 328: Are you sure at this point in the code that `s.SegmentStoreResources.Limits` exists?

pkg/test/e2e/e2eutil/pravegacluster_util.go
```diff
@@ -159,35 +159,34 @@ func CreatePravegaClusterWithTlsAuth(t *testing.T, f *framework.Framework, ctx *
...
-		"autoScale.tlsCertFile":                   "/etc/secret-volume/ca-cert",
-		"bookkeeper.tlsEnabled":                   "true",
-		"bookkeeper.tlsTrustStorePath":            "empty",
-		"autoScale.validateHostName":              "true",
-		"autoScale.authEnabled":                   "true",
-		"controller.auth.tokenSigningKey":         "secret",
-		"autoScale.tokenSigningKey":               "secret",
-		"pravega.client.auth.token":               "YWRtaW46MTExMV9hYWFh",
-		"pravega.client.auth.method":              "Basic",
+		"pravegaservice.container.count":                                    "4",
```
RaulGracia, line 162: Good, this was still with the old nomenclature,

RaulGracia (approved):
LGTM, just one comment.

pkg/apis/pravega/v1beta1/pravega.go
nishant-yt: yes because in `ValidateSegmentStoreMemorySettings` method we are first checking whether `p.Spec.Pravega.SegmentStoreResources.Limits == nil` and in turn making sure user does pass Limits

anishakj (approved):
LGTM

PR #598, opened 2021-12-03
Adding validation checks for Bookkeeper Deployments

### LINKED ISSUE #587: Adding a values check for 1-BK deployments
### Description

Verify whether the bookkeeper cluster has been deployed with one replica and if so , make sure the following condition is met `bookkeeper.ensemble.size >=  bookkeeper.write.quorum.size  >= bookkeeper.ack.quorum.size`

### Importance

must-have

### Location

https://github.com/pravega/pravega-operator/blob/master/pkg/apis/pravega/v1beta1/pravegacluster_types.go

The following checks should be performed :

1. bookkeeper.ensemble.size >=  bookkeeper.write.quorum.size  >= bookkeeper.ack.quorum.size
2. If bookkeeper.ensemble.size == 1 then `bookkeeper.write.quorum.racks.minimumCount.enable` should be set to `false`

nishant-yt:
Signed-off-by: Nishant Gupta <Nishant_Gupta3@dell.com>

### Change log description

When deploying the pravega cluster,  the options `bookkeeper.ensemble.size`  `bookkeeper.write.quorum.size`  `bookkeeper.ack.quorum.size` should be set such that the following condition is satisfied: `bookkeeper.ensemble.size >= bookkeeper.write.quorum.size >= bookkeeper.ack.quorum.size`

### Purpose of the change

Fixes #587

### What the code does

The code preforms the following checks :
- `bookkeeper.ensemble.size >=  bookkeeper.write.quorum.size  >= bookkeeper.ack.quorum.size`
- If `bookkeeper.ensemble.size == 1` then `bookkeeper.write.quorum.racks.minimumCount.enable` should be set to `false`

### How to verify it

Install a pravega cluster with different values of  `bookkeeper.ensemble.size`  `bookkeeper.write.quorum.size`  `bookkeeper.ack.quorum.size` and in case the above mentioned condition in not met , then an error is retuned by the validating webhook. 
In case the values adheres to the above condition say  `bookkeeper.ensemble.size=4`  `bookkeeper.write.quorum.size=3`  `bookkeeper.ack.quorum.size=3` then post installation of the pravega cluster if you go inside Controller/ Segment Store Pod then all these values should be available in the env variable `JAVA_OPTS` as

>  JAVA_OPTS=-Dbookkeeper.ack.quorum.size=3 -Dbookkeeper.ensemble.size=4 -Dbookkeeper.write.outstanding.bytes.max=33554432 -Dbookkeeper.write.quorum.size=3 -Dbookkeeper.write.timeout.milliseconds=60000 -Dcontroller.container.count=8 -Dcontroller.retention.bucket.count=4 -Dcontroller.retention.thread.count=4 -Dcontroller.service.asyncTaskPool.size=20

### DIFF
--- a/pkg/apis/pravega/v1beta1/pravega.go
+++ b/pkg/apis/pravega/v1beta1/pravega.go
@@ -383,6 +383,21 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
 		s.Options = map[string]string{}
 	}
 
+	if s.Options["bookkeeper.ensemble.size"] == "" {
+		changed = true
+		s.Options["bookkeeper.ensemble.size"] = "3"
+	}
+
+	if s.Options["bookkeeper.write.quorum.size"] == "" {
+		changed = true
+		s.Options["bookkeeper.write.quorum.size"] = "3"
+	}
+
+	if s.Options["bookkeeper.ack.quorum.size"] == "" {
+		changed = true
+		s.Options["bookkeeper.ack.quorum.size"] = "3"
+	}
+
 	if s.ControllerJvmOptions == nil {
 		changed = true
 		s.ControllerJvmOptions = []string{}
--- a/pkg/apis/pravega/v1beta1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1beta1/pravegacluster_types.go
@@ -905,6 +905,10 @@ func (p *PravegaCluster) ValidateCreate() error {
 	if err != nil {
 		return err
 	}
+	err = p.ValidateBookkeperSettings()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -920,6 +924,10 @@ func (p *PravegaCluster) ValidateUpdate(old runtime.Object) error {
 		return err
 	}
 	err = p.ValidateSegmentStoreMemorySettings()
+	if err != nil {
+		return err
+	}
+	err = p.ValidateBookkeperSettings()
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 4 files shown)


pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1296,6 +1304,72 @@ func (p *PravegaCluster) ValidateSegmentStoreMemorySettings() error {
...
+		return fmt.Errorf("bookkeeper.write.quorum.racks.minimumCount.enable can be only set to \"true\" \"false\" or \"\"")
+	}
+
+	if writeQuorumRacks == "true" && ensembleSizeInt == 1 {
+		return fmt.Errorf("bookkeeper.write.quorum.racks.minimumCount.enable should be set to false if bookkeeper.ensemble.size is 1")
+	}
+
+	if ensembleSizeInt < writeQuorumSizeInt {
+		if ensembleSize == "" {
+			return fmt.Errorf("The value provided for the option bookkeeper.write.quorum.size should be less than or equal to the default value of option bookkeeper.ensemble.size which is 3")
```
jkhalack, line 1352: The message is a bit confusing (the user might have forgotten to set `bookkeeper.ensemble.size` altogether).
Maybe change it to something like this: 
`"The value provided for the option bookkeeper.write.quorum.size should be less than or equal to the value of option bookkeeper.ensemble.size (default is 3)"`
Then the user would be aware of the need to **either** set the ensemble size explicitly **or** coordinate the quorum size with the default.

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1296,6 +1304,72 @@ func (p *PravegaCluster) ValidateSegmentStoreMemorySettings() error {
...
+	}
+
+	if writeQuorumSizeInt < ackQuorumSizeInt {
+		if writeQuorumSize == "" {
+			return fmt.Errorf("The value provided for the option bookkeeper.ack.quorum.size should be less than or equal to the default value of option bookkeeper.write.quorum.size which is 3")
+		}
+		if ackQuorumSize == "" {
+			return fmt.Errorf("The value provided for the option bookkeeper.write.quorum.size should be greater than or equal to the default value of bookkeeper.ack.quorum.size which is 3")
+		}
+		return fmt.Errorf("The value provided for the option bookkeeper.ack.quorum.size should less than or equal to the value of option bookkeeper.write.quorum.size")
```
jkhalack, line 1367: (nit) `should less than` -> `should be less than`

jkhalack (approved):

pkg/apis/pravega/v1beta1/pravega.go
```diff
@@ -383,6 +383,21 @@ func (s *PravegaSpec) withDefaults() (changed bool) {
 		s.Options = map[string]string{}
 	}
 
+	if s.Options["bookkeeper.ensemble.size"] == "" {
```
anishakj, line 386: @nishant-yt  since we are making all the 3 fields mandatory, will it enter this code flow

nishant-yt: @anishakj all the fields are not mandatory. If the user doesn't provide value for any of these options, we are setting it to `3` by default. 
One of instances when it will enter the above mentioned code flow is when the user doesn't provide any value for `bookkeeper.ensemble.size` `bookkeeper.write.quorum.size` and `bookkeeper.ack.quorum.size`

anishakj (approved):
LGTM

PR #612, opened 2022-03-03
Issue 611: Adding validation for authentication settings

### LINKED ISSUE #611: Enhance validation for authentication settings
### Description

Currently, if user has enabled authentication via `spec.Authentication.Enabled` but not provided correct options, Pravega will get deployed  and it fails to list streams

### Importance

 must-have

### Location

pkg/apis/pravega/v1beta1/pravegacluster_types.go

Add webhook validation to ensure authentication options are correct

anishakj:

### Change log description

Adding validation for authentication settings

### Purpose of the change

Fixes #611
### What the code does

If authentication is enabled,  webhook validation will  ensure
a. Pravega.Options["autoScale.controller.connect.security.auth.enable"]/p.Spec.Pravega.Options["autoScale.authEnabled"] is set to true
b. controller.security.auth.delegationToken.signingKey.basis is present
c. autoScale.security.auth.token.signingKey.basis is present
d. b and c are equal

If authentication is disabled. validation will ensure  Pravega.Options["autoScale.controller.connect.security.auth.enable"]/p.Spec.Pravega.Options["autoScale.authEnabled"] is set to false
### How to verify it

Verified by setting these options and installed pravega and saw that proper error message is displayed.

### DIFF
--- a/pkg/apis/pravega/v1beta1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1beta1/pravegacluster_types.go
@@ -1268,12 +1268,10 @@ func (p *PravegaCluster) ValidateAuthenticationSettings() error {
 			return fmt.Errorf("controller and segmentstore token signing key should have same value")
 		}
 	} else {
-		newkey, ok := p.Spec.Pravega.Options["autoScale.controller.connect.security.auth.enable"]
-		oldkey := ""
-		if !ok {
-			oldkey, ok = p.Spec.Pravega.Options["autoScale.authEnabled"]
-		}
-		if ok && (oldkey == "true" || newkey == "true") {
+		newkey, _ := p.Spec.Pravega.Options["autoScale.controller.connect.security.auth.enable"]
+		oldkey, _ := p.Spec.Pravega.Options["autoScale.authEnabled"]
+
+		if oldkey == "true" || newkey == "true" {
 			return fmt.Errorf("autoScale.controller.connect.security.auth.enable/autoScale.authEnabled should not be set to true")
 		}
 	}
--- a/pkg/apis/pravega/v1beta1/pravegacluster_types.go
+++ b/pkg/apis/pravega/v1beta1/pravegacluster_types.go
@@ -911,6 +911,11 @@ func (p *PravegaCluster) ValidateCreate() error {
 	if err != nil {
 		return err
 	}
+
+	err = p.ValidateAuthenticationSettings()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -930,6 +935,10 @@ func (p *PravegaCluster) ValidateUpdate(old runtime.Object) error {
 		return err
 	}
 	err = p.ValidateBookkeperSettings()
+	if err != nil {
+		return err
+	}
+	err = p.ValidateAuthenticationSettings()
 	if err != nil {
 		return err
 	}
@@ -1000,6 +1009,7 @@ func (p *PravegaCluster) ValidatePravegaVersion() error {
 	return nil
 }
 
+//
 func (p *PravegaCluster) validateConfigMap() error {
 	configmap := &corev1.ConfigMap{}
 	err := Mgr.GetClient().Get(context.TODO(),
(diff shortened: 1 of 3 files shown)


pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1000,6 +1009,7 @@ func (p *PravegaCluster) ValidatePravegaVersion() error {
 	return nil
 }
 
+//
```
jkhalack, line 1012: Left over from some other commit?

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1216,6 +1226,50 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+func (p *PravegaCluster) ValidateAuthenticationSettings() error {
+	if p.Spec.Authentication.Enabled == true {
+		newkey, ok := p.Spec.Pravega.Options["autoScale.controller.connect.security.auth.enable"]
+		oldkey := ""
+		if !ok {
+			oldkey, ok = p.Spec.Pravega.Options["autoScale.authEnabled"]
+		}
+		if !ok {
+			return fmt.Errorf("autoScale.controller.connect.security.auth.enable field is not present")
+		} else if newkey == "false" || oldkey == "false" {
```
jkhalack, line 1240: Should we rather be checking instead that one (or both) of them is set to `"true"` and the other is unset (or both are set to `"true"`)?

anishakj: I am not sure if one of the keys is set to `true` and other is set to `false` which one will take precedence. I think second option will take precedence. But here .we are ensuring that both of the keys cannot be set to `false`.

jkhalack: My concern was more about things like `"True"` or `"dummy"` (there is no validation)

anishakj: Made the changes to take care of that scenario and verified it. Please have a look

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1216,6 +1225,59 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+			}
+		}
+		if oldkey == "" {
+			if newkey != "true" {
+				return fmt.Errorf("autoScale.controller.connect.security.auth.enable should be set to true")
+			}
+		}
+		if oldkey != "" && newkey != "" {
+			if newkey != "true" || oldkey != "true" {
+				return fmt.Errorf("autoScale.controller.connect.security.auth.enable/autoScale.authEnabled should be set to true")
```
nishant-yt, line 1250: Since at this point user is setting both the Options , we can modify the Error message as:
`Both autoScale.controller.connect.security.auth.enable and autoScale.authEnabled should be set to true`

nishant-yt (approved):
LGTM

pkg/apis/pravega/v1beta1/pravegacluster_types.go
```diff
@@ -1216,6 +1225,59 @@ func (p *PravegaCluster) validateConfigMap() error {
...
+		if oldkey == "" {
+			if newkey != "true" {
+				return fmt.Errorf("autoScale.controller.connect.security.auth.enable should be set to true")
+			}
+		}
+		if oldkey != "" && newkey != "" {
+			if newkey != "true" || oldkey != "true" {
+				return fmt.Errorf("Both autoScale.controller.connect.security.auth.enable and autoScale.authEnabled should be set to true")
+			}
+		}
```
jkhalack, line 1252: The logic here won't allow disabling the authentication, which does not seem to be the purpose of the current PR.

jkhalack (changes requested):

pkg/apis/pravega/v1beta1/pravegacluster_types.go
anishakj: @jkhalack  we can disable authentication by making ```authentication:
  enabled: false```  and in that case we are just ensuring option fields are not set to `true`

jkhalack: We are not getting to line 1270 unless we set either one or both of `autoScale.authEnabled`, `autoScale.controller.connect.security.auth.enable` keys to something.
None of the above keys can be set to `false` and allow the logic slip to line 1270.

PR #619, opened 2022-04-25
Issue 618: Adding support to disable finalizer

### LINKED ISSUE #618: Operator is not starting when disable finalizer flag is set
### Description

Operator fails to start if disable finalizer flag is set

### Importance

 must-have

### Location

main.go

Provide support to disable finalizer

anishakj:

### Change log description

Added flag disable finalizer
### Purpose of the change

Fixes #618

### What the code does

Added disable finalizer flag.
passing namespace when manager is starting.
Calling customized leader.Become function 

### How to verify it

All e2e should pass

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,17 +18,22 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


nishant-yt (approved):
LGTM

PR #621, opened 2022-05-09
Issue 620: Fix for pravega image upgrade failure

### LINKED ISSUE #620: Upgrade of pravega image is failing in certain cases
### Description

Install Pravega operator and pravega  cluster. Update the pravega image along with that add more labels to pravega pods. Upgrade is failing with error
```
time="2022-05-04T16:16:47Z" level=info msg="error syncing segmentstore component version, upgrade failed. could not obtain outdated pod"
```
### Importance

must-have

### Location

upgrade.go

Fix the upgrade issue

anishakj:

### Change log description

Made sure that correct labels are used for selecting pods during upgrade
### Purpose of the change

Fixes #620
### What the code does

During upgrade process, if the labels are also changed, while listing pods it is looking for pods with updated labels and which results in pods are not getting listed. Corrected the logic for finding labels

In order  to fix this we are using a subset of labels that will be present always.  labels used in selector are
``` 
"app": "pravega-cluster",
"pravega_cluster": pravegaCluster.Name,
```

### How to verify it

Verified that upgrade is working fine even with the change of labels.

### DIFF
--- a/controllers/upgrade.go
+++ b/controllers/upgrade.go
@@ -399,7 +399,8 @@ func (r *PravegaClusterReconciler) syncSegmentStoreVersion(p *pravegav1beta1.Pra
 	}
 
 	if ready {
-		pod, err := r.getOneOutdatedPod(sts, p.Status.TargetVersion)
+		labels := p.LabelsForPravegaCluster()
+		pod, err := r.getOneOutdatedPod(sts, p.Status.TargetVersion, labels)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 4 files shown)


controllers/upgrade.go
```diff
@@ -399,7 +399,8 @@ func (r *PravegaClusterReconciler) syncSegmentStoreVersion(p *pravegav1beta1.Pra
 	}
 
 	if ready {
-		pod, err := r.getOneOutdatedPod(sts, p.Status.TargetVersion)
+		labels := p.LabelsForPravegaCluster()
```
jkhalack, line 402: Would this function fetch new or old labels?

anishakj: `LabelsForPravegaCluster()` is a subset of labels and that will be present  as mandatory labels in every sts.

nishant-yt (approved):
LGTM

PR #624, opened 2022-05-16
Issue 623: Remove version map

### LINKED ISSUE #623: Remove the version map
### Description

Currently pravega-operator is not using version map in the operator code. Remove the version map from manifests and update the documentation as needed.

### Importance
 must-have

### Location

deploy folder

Remove the version map yaml

anishakj:
### Change log description

Removed the supported version map manifests as it is  not used in operator. Updated `operator.yaml` to remove the volume mounts of version map. Removed the deploy folder completely as it is already present in config folder.

### Purpose of the change

 Fixes #623

### What the code does

Removed the version map and removed the unused yaml files. Updated `operator.yaml`  to not use version map.

### How to verify it

All e2e should pass


config/webhook/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+  - webhook.yaml
```
nishant-yt, line 2: can we remove the file "`manifests.yaml`" from **config/webhook** folder in case it is not used anywhere

anishakj: sure

nishant-yt (approved):
GOOD TO GO

PR #662, opened 2023-04-14
Issue 661: Removing the check of controller minimum replicas from operator

### LINKED ISSUE #661: Remove the check of minimum number of controller replicas for operator
### Description

Remove the check for minimum number of controller replicas
### Importance
should-have

### Location

api/v1beta1/pravega.go

Provide the required support

anishakj:
### Change log description

Adding the support additional port in segment store service

### Purpose of the change

 Fixes #661

### What the code does
Removed the check of mandating controller replica to 1
### How to verify it
Removed the check of mandating controller replica to 1
currently the check for minimum controller replicas is removed, because during LTS recovery we need to stop all the requests from controller. In order to achieve this we are allowing `0` controller replicas.

### DIFF
--- a/api/v1beta1/pravega.go
+++ b/api/v1beta1/pravega.go
@@ -363,11 +363,6 @@ type Probe struct {
 }
 
 func (s *PravegaSpec) withDefaults() (changed bool) {
-	if !config.TestMode && s.ControllerReplicas < 1 {
-		changed = true
-		s.ControllerReplicas = 1
-	}
-
 	if s.RollbackTimeout < 1 {
 		changed = true
 		s.RollbackTimeout = 10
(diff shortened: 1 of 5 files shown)


api/v1beta1/pravega.go
```diff
@@ -363,11 +363,6 @@ type Probe struct {
 }
 
 func (s *PravegaSpec) withDefaults() (changed bool) {
-	if !config.TestMode && s.ControllerReplicas < 1 {
```
jkhalack, line 366: Is there a reason to stop checking for negative controller replicas?

anishakj: Yes @jkhalack there is a requirement for LTS recovery

jkhalack (approved):

api/v1beta1/pravega.go
shshashwat: what if we get negative number here?

anishakj: @shshashwat we are having openapi validation, that wont allow negative numbers https://github.com/pravega/pravega-operator/blob/master/api/v1beta1/pravega.go#L157

krishan-Rai (approved):
LGTM

PR #671, opened 2023-09-11
Issue 670: Updating the segmentstore port

### LINKED ISSUE #670: Segment store pods are not coming up in K3s cluster
### Description
While installing pravega in K3s cluster, segement store pods are not coming up with the below error 

External IPs are in pending state for services

### Importance

 must-have

### Location

pravega_segmentstore.go

Fix the port issue

anishakj:
### Change log description

Updating the segment store service port incrementally if  `SegmentStoreLoadBalancerIP` is set

### Purpose of the change

 Fixes #670

### What the code does
If we are using same segment store service IP with different ports, in case of admin port as well port number needs to be incremented. If it is not incremented will give port conflict

### How to verify it
Verify segment store pods are coming up fine

### DIFF
--- a/controllers/pravega_segmentstore.go
+++ b/controllers/pravega_segmentstore.go
@@ -641,6 +641,7 @@ func MakeSegmentStoreExternalServices(p *api.PravegaCluster) []*corev1.Service {
 		}
 		if p.Spec.Pravega.SegmentStoreLoadBalancerIP != "" {
 			service.Spec.Ports[0].Port = int32(serviceport) + i
+			service.Spec.Ports[1].Port = int32(adminPort) + i
 			service.Spec.LoadBalancerIP = p.Spec.Pravega.SegmentStoreLoadBalancerIP
 		}
 		services[i] = service


controllers/pravega_segmentstore.go
```diff
@@ -641,6 +641,7 @@ func MakeSegmentStoreExternalServices(p *api.PravegaCluster) []*corev1.Service {
 		}
 		if p.Spec.Pravega.SegmentStoreLoadBalancerIP != "" {
 			service.Spec.Ports[0].Port = int32(serviceport) + i
+			service.Spec.Ports[1].Port = int32(adminPort) + i
```
jkhalack, line 644: @anishakj is it expected to go up (9999, 1000, 1001) or down (9999, 9998, 9997)?

anishakj: ports are supported upto `65535`

krishan-Rai (approved):
LGTM

PR #672, opened 2023-10-09
Issue 643: Fixing intermittent upgrade failure

### LINKED ISSUE #643: pravega cluster upgrade is failing intermittently

anishakj:
### Change log description

Fixing intermittent upgrade failure of pravega.
Added some more conditions to handle error scenarios

### Purpose of the change

Fixes #643

### What the code does

while during segmentstore update, it fails to get the pod sometimes. Issue is happening because while it tries to get an outdated pod, it has already updated. So added more checks to post error only if all pods were not updated

### How to verify it
Peformed upgrade multiple time and issue is not seen

### DIFF
--- a/controllers/upgrade.go
+++ b/controllers/upgrade.go
@@ -414,7 +414,7 @@ func (r *PravegaClusterReconciler) syncSegmentStoreVersion(p *pravegav1beta1.Pra
 				return false, err
 			}
 			if *sts.Spec.Replicas == (int32)(len(pods)) {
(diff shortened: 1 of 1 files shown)


jkhalack (approved):

krishan-Rai (approved):
LGTM

PR #675, opened 2023-11-07
Issue 674: Fix for segment store services are not coming up properly

### LINKED ISSUE #674: Segment store services are not coming up properly
### Description

While installing pravega with 12 segment store pods , installation is getting stuck

### Importance

must-have

### Location

controllers/pravega_segmentstore.go
api/v1beta1/pravegacluster_types.go

Make sure segmentstore services are not using any of the port used by the cluster for other purposes

anishakj:
### Change log description

While deploying pravega with 12 segmentstore pods, segmentstore service was using one of the ports used by other service `containerd` and was causing conflict
Added new spec field in CRD as `reservedPortList` and while assigning port to service ensure that it is not in reserved port List 

### Purpose of the change

Fixes #674

### What the code does

Assigning port to segment store services which is not in the reserved portList, if it finds conflict next  port will be assigned.

### How to verify it

Verified pravega installation with 12 segmentstore pods and ensured that ports mentioned in reservedportlist are getting skipped.

### DIFF
--- a/controllers/pravega_segmentstore.go
+++ b/controllers/pravega_segmentstore.go
@@ -594,8 +594,8 @@ func MakeSegmentStoreExternalServices(p *api.PravegaCluster) []*corev1.Service {
 	serviceport, _ := strconv.Atoi(p.Spec.Pravega.Options["pravegaservice.service.listener.port"])
 	adminPort, _ := strconv.Atoi(p.Spec.Pravega.Options["pravegaservice.admin.listener.port"])
 
-	servicePortIncrement := int32(0)
-	adminPortIncrement := int32(0)
+	currentServicePort := int32(serviceport)
+	currentAdminPort := int32(adminPort)
 
 	for i := int32(0); i < p.Spec.Pravega.SegmentStoreReplicas; i++ {
 		ssPodName := p.ServiceNameForSegmentStore(i)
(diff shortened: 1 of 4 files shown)


controllers/pravega_segmentstore.go
```diff
@@ -640,8 +644,17 @@ func MakeSegmentStoreExternalServices(p *api.PravegaCluster) []*corev1.Service {
...
+				servicePortIncrement++
+			}
+
+			service.Spec.Ports[0].Port = int32(serviceport) + i + servicePortIncrement
+
+			for util.ContainsElement(p.Spec.ReservedPortList, int32(adminPort)+i+adminPortIncrement) {
+				adminPortIncrement++
+			}
+			service.Spec.Ports[1].Port = int32(adminPort) + i + adminPortIncrement
+
```
jkhalack, line 657: I'm not sure why we have all this math here.
We could have a `currentServicePort` initialized to `int32(serviceport)` and a `currentAdminPort` initialized to `int32(adminPort)`. 
At the end of every iteration of SS replicas we run `currentServicePort++` and `currentAdminPort++`.
 If `currentServicePort` value is on reserved list, we run `currentServicePort++`.
Again, if `currentAdminPort` value is on reserved list, we run  `currentAdminPort++`

anishakj: Thanks @jkhalack for the feedback, have made changes as suggested.

jkhalack (approved):
LGTM

krishan-Rai (approved):
LGTM

